PR #89, opened 2016-05-16
Tatami mat

yishn:
Screenshot

- The board size will adapt to the grid size
- This also fixes rounding problems with the hoshi points 
- Current stone markers resizes with board without having rounding problems

Non-square boards


fohristiwhirl:
This seems to waste screen space? Currently the board always goes more or less to the edge of the window.

yishn:
It's really not. The stone size stays the same with the same window size.

yishn:
It's actually not. The stone size stays the same with the same window size. This only adapts the board size to what's actually needed and fills the superfluous empty space with a background image.

fohristiwhirl:
OK fair enough. (I mean personally I prefer the current way where the added background is just the same wood texture, but perhaps I'll be able to make that happen via userstyle...)

Issue #121, opened 2016-08-24
Support non-UTF8 encodings


GreatHX:
Thanks for doing such a cool Go software! 
But when I am trying to open SGF files with Chinese comments in them, the text doesn't display very well(Mac Version). 

<img width="789" alt="2016-08-24 13 40 49" src="https://cloud.githubusercontent.com/assets/21212927/17919469/67b4cee0-6a00-11e6-8093-a8f6460014c2.png">

yishn:
Thanks! I'm currently on vacation, but I'll investigate when I come back.

apetresc:
@GreatHX It's likely an encoding issue – I tried with some UTF-8 encoded Chinese characters and it worked great. Can you perhaps attach the SGF you're trying to open to this ticket? (You can do that just by dragging-and-dropping it onto the comment box). That will help diagnose the problem.

GreatHX:
@yishn @apetresc Thanks！This is the SGF with the problem.

[围棋定式一月通.sgf.zip](https://github.com/yishn/Sabaki/files/435859/sgf.zip)

apetresc:
Yes, indeed, the problem is that this file is encoded with GB18030 which is not being detected properly. Converting it to UTF-8 with, say:

```
iconv -f GB18030 -t UTF-8 chinese.sgf > chinese_utf8.sgf
```
results in a readable file (attached):
![screenshot 2016-08-24 21 07 03](https://cloud.githubusercontent.com/assets/14872/17953126/b78ab43c-6a3e-11e6-88c2-594c7d21eefa.png)

Given that the encoding was properly set in the SGF (through the `CA` property) it should be possible to detect and apply this. I'll take a look and see if it's easy enough for me to do it :)
[chinese_utf8.sgf.zip](https://github.com/yishn/Sabaki/files/435960/chinese_utf8.sgf.zip)

GreatHX:
@apetresc Thanks a lot for your answer!

apetresc:
@GreatHX I've implemented a fix in my branch [apetresc/sabaki@encoding](https://github.com/apetresc/sabaki/tree/encoding) which works for your file. There's still a bit more work to be done, though, and I'd love to get @yishn's thoughts on it before I fix the last few bits.

GreatHX:
@apetresc I really appreciate your fixing!

solatale:
[...]

PR #122, opened 2016-08-25
Support non-UTF8 encodings

apetresc:
Status
=====
 - [x] Implement support for reading non-UTF-8 files
 - [x] Add tests for Chinese SGFs
 - [x] Add support for saving files back to their *non*-UTF-8 forms (if this is even desirable).
 - [x] More tests with more languages/encodings, and some unsupported encodings as well.

Screenshots
=========
screenshot 2016-08-24 23 47 47
(The exciting part of this screenshot is not that Chinese characters are being displayed correctly, since Sabaki can already do that; the exciting part is that the file being opened is not UTF-8 encoded, as can be seen from the terminal).

Known Issues
==========
 - [x] Currently, if the `CA[]` property comes *after* a text node in the root gametree, the text node preceding it won't be properly decoded if it's non-ISO.

Addresses yishn/sabaki#121.

*(Note: this is my first time writing any Node/Electron at all, so this may not be very idiomatic 😄)*


apetresc:
On further reflection, I think the only safe way to set the encoding is to make two passes: one where we we just parse every field as ISO-8859-1 in order to read the `CA` attribute (if present), and then if it is set, another parsing round where text fields are read properly.

I think this is the cleanest way to get around the problem that the parser doesn't know how to interpret text attributes that occur _earlier_ than the `CA` tag. What do you think, @yishn?

yishn:
Thank you so much for your work @apetresc! I'll look over it once I'm back.

yishn:
Yes, it's probably the best way. If the CA tag only appears in the root node, we only need to parse just a few tokens to get to the CA property.

yishn:
Another idea is as follows, since the CA property can only appear in the root node: With the current implementation, the only properties not handled are the properties in the parent node of the CA property. A quick (and maybe dirty) way is to just fix said properties when we encounter a CA tag. This would also work if there are several games with different CA properties inside one sgf file.

yishn:
I'm back from vacation! @apetresc do you want to finish this PR, or should I take over?

apetresc:
I can finish it! I'll implement your suggestion on Monday.

How do you feel about the saving aspect? Do you want to preserve the `CA[]` tag and encode with it when saving, or convert everything to `UTF-8`?

yishn:
I don't really know, but personally, I'm fine with saving everything in UTF-8.

apetresc:
@yishn, I think I'm done! I've completed all the objectives set out initially, and it works well from my personal manual testing, including saving back to UTF-8.

Let me know if there's anything else that needs doing :)

modules/sgf.js
```diff
@@ -10,11 +10,21 @@ if (typeof require != 'undefined') {
...
 }
 
 var context = typeof module != 'undefined' ? module.exports : (window.sgf = {})
 var alpha = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'
 
+/* The default encoding and list of properties that should be interpreted as
+ * being encoded by the file's CA[] property is defined in the SGF spec at
+ * http://www.red-bean.com/sgf/properties.html#CA
+ */
+var default_encoding = 'ISO-8859-1'
```
yishn: Can you change the variable names to camelCase style to fit with convention?

package.json
```diff
@@ -16,9 +16,11 @@
   "dependencies": {
     "argv-split": "^1.0.0",
     "gemini-scrollbar": "^1.4.3",
+    "iconv-lite": "^0.4.13",
     "octicons": "^4.3.0",
     "pikaday": "^1.4.0",
-    "sigma": "^1.0.3"
+    "sigma": "^1.0.3",
+    "tmp": "0.0.28"
```
yishn: Shouldn't `tmp` be in devDependencies if it's only needed in the tests?

yishn:
Thanks for your hard work! I left some comments in the code. I have one question, though: 

In [case we have an encoding and try to parse values](https://github.com/yishn/Sabaki/pull/122/files#diff-7183a0b6f43c8f1ab8ceef19d6bb450cR113) we first decode the encoded data, then apply unescapeString, i.e. `unescapeString(decode(encoded_data))`, however when we [encounter `CA` and reparse values](https://github.com/yishn/Sabaki/pull/122/files#diff-7183a0b6f43c8f1ab8ceef19d6bb450cR98), we actually do `decode(unescapeString(encoded_data))`. I'm wondering if this yields the same thing?

I have no experience with text encodings and I'm not sure what the "correct" order is, but they seem to give the same results.

modules/sgf.js
apetresc: Whoops, not sure how I missed that! Done :)

package.json
apetresc: Definitely – I just didn't know `devDependencies` existed, this is my first time touching Node. Done!

apetresc:
Excellent point about the different order of escaping. In theory, it *could* make a difference if the encoding we're switching between had a different byte for `"` or `\` characters. In practice, though, virtually all 8-bit encodings in common use today (and probably all the ones that `iconv-lite` actually supports) are ASCII-compatible and don't differ on those common characters.

yishn:
Thank you for clarifying! I'm merging this now.

Issue #158, opened 2017-01-09
Generate a .desktop file for Linux builds


apetresc:
Most Linux desktops can only associate file extensions with [.desktop files](https://wiki.archlinux.org/index.php/Desktop_entries), not random binaries. So, without one, you can't associate Sabaki with SGFs. Creating one yourself isn't a big deal, but a lot of applications will just generate one for you (and `make install` will put it into `/usr/share/applications` but I guess Sabaki isn't supposed to have an install script). So it might be nice if this was part of the `npm run build:linux` flow.

apetresc:
Hmm, I was trying to implement this one myself, but I can't seem to find the `electron-builder` configuration at all, only the `electron-packager`. Is the `electron-builder` configuration not committed to the repository to protect the signing keys, etc, or do you just not use it at all?

yishn:
Currently, I only use electron-packager. Is there a reason you need electron-builder?

apetresc:
Oh, no, I just assumed you were using `electron-builder` since the Mac/Windows builds prompt to auto-update, which I thought was an `electron-builder`-only feature.

In that case, I'm not sure if `electron-packager` even knows how to generated `.desktop` files by itself, might have to just manually render a template =/

yishn:
Haha... sadly, it's not a prompt to auto-update, it's a prompt to manually update. I just check the GitHub releases page if the version is up to date. 😆 It works under Linux as well.

apetresc:
Ah, okay :) I just build from source, so I never used the auto-update, I
assumed it actually ran it!

So, in that case, maybe I can rephrase this ticket as - generate
installers/packages for each platform using `electron-builder`. Seems like
it will have even more utility than just Linux desktop integration. It's
also something I can try to tackle, if you'd like.

What do you think?

yishn:
I've looked into `electron-builder`, it seems to have a lot of options... If you want to tackle it, then I'll give my OK. 👍

PR #165, opened 2017-02-06
Switch to electron-builder

apetresc:
This addresses #158.

Electron-packager generates standalone packages, while electron-builder can generate platform-specific installers. There are several benefits of electron-builder, including:

* Generate familiar-feeling installers for each platform: .DMG images for OS X, MSI installers for Windows (which can be installed/uninstalled from the system Control Panel), and .deb/.rpm/PKGBUILD files for Linux.
* Nicer configuration in `package.json` in my opinion. There's a separate config section instead of passing in a bunch of args to each platform's `build` step.
* SGF file association at the OS level.
* Integration with native app stores, if desired (this isn't part of this PR obviously, but if you ever wanted to be present on the MacOS app store, the Ubuntu repositories, etc, then this is necessary)
* Support for auto-updates (also not part of this PR since it requires some setup on the GitHub side first)

Some major things I need help with:

* Some package managers require icon files that are larger than the 256x256 icon that's currently on GitHub. As a temporary workaround, I manually enlarged the 256x256 to 1024x1024 and let electron-builder scale it back down so that the build passes. So the icon looks a bit ugly right now. Obviously it would be ideal to get a native 1024x1024 version... hopefully you still have the vector image :)
* To get auto-updates working, you need to have one full release on GitHub through electron-builder, and then you have to set up some feeds.

I've tested on all three major platforms on my machines, and everything seems to work well, as far as I can tell. Let me know if you prefer these commits squashed!


package.json
```diff
@@ -3,12 +3,40 @@
...
+    },
+    "fileAssociations": [
+      {
+        "ext": "sgf",
+        "name": "SGF"
+      }
+    ],
+    "files": [
+      "**/*",
+      "!(build.js|docs|tests?|examples|.c9|scss|^/node_modules/(moment|(octicons|sigma)/(plugins|lib|src|build/font)))($|/)"
```
apetresc: Ideally we can now split this massive rule up into a series of smaller rules, since `build.files` supports that. I just didn't wanna mess with it until I was sure it was correct first.

yishn:
Thank you so much! I'll look over it after my exams.

package.json
```diff
@@ -3,12 +3,40 @@
...
+      "category": "Game",
+      "packageCategory": "games",
+      "target": ["deb", "rpm", "pacman", "tar.gz"]
+    },
+    "mac": {
+      "category": "public.app-category.board-games"
+    },
+    "win": {
+      "iconUrl": "https://raw.githubusercontent.com/yishn/sabaki/master/data/icon.ico",
+      "target": ["msi"]
```
yishn: Is there a `msi` target for Windows? It doesn't work on my machine and I couldn't find the target in the [documentation](https://github.com/electron-userland/electron-builder/wiki/Options#WinBuildOptions-target). Maybe we can just provide a 7z or use NSIS?

package.json
```diff
@@ -3,12 +3,40 @@
...
+    },
+    "fileAssociations": [
+      {
+        "ext": "sgf",
+        "name": "SGF"
+      }
```
yishn: Does this mean `build/extra-info.plist` is redundant now?

package.json
yishn: Looking at the [documentation](https://github.com/electron-userland/electron-builder/wiki/Options#file-patterns), it seems like electron-builder isn't using Regex, but glob patterns. I'll look into it...

Issue #160, opened 2017-01-14
Copying to and loading from clipboard for web build


C0DEHERO:
The web build doesn't seem to support copying the sgf to the clipboard or loading an sgf from the clipboard, even though that would be very convenient feature for the web version. Since the desktop build already has that feature, it probably shouldn't be too hard to add it to the web build.

yishn:
Copying things to the clipboard is not a problem nowadays, but AFAIK the browser can't read the clipboard.

C0DEHERO:
Well no, but you could pop up a text box to paste the sgf inside, sort of like the gokibitz text upload box.

yishn:
Yeah, I just wanted to point out that the implication "desktop build already has that feature ⟹ it shouldn't be too hard" is not correct. 😃

C0DEHERO:
Oh, true I guess :D I just figured you wouldn't have to write much extra code since the desktop and web build use the same code base, but then I don't know anything about javascript or electron.

PR #167, opened 2017-02-17
web version clipboard loading and saving

C0DEHERO:
WIP PR to implement the features suggested in issue #160.

The code in its current state works perfectly with Ctrl+C and Ctrl+V and with the "Save to Clipboard" menu option. I am not sure if it is possible to implement the "Load from Clipboard" feature in a functional manner, so it may be necessary to add a text popup, which the user can manually copy. Or perhaps the copy/paste shortcuts would be enough with a instructions in the UI on how to use them.

### DIFF
--- a/browser/index.html
+++ b/browser/index.html
@@ -16,7 +16,6 @@
 <script src="https://cdn.rawgit.com/CamHenlin/TouchPolyfill/master/touchpolyfill.min.js#sabaki-v0.11.0"></script>
 <script src="../node_modules/sigma/build/sigma.min.js"></script>
 <script src="../node_modules/sigma/build/plugins/sigma.renderers.customShapes.min.js"></script>
-<script src="../node_modules/clipboard/dist/clipboard.min.js"></script>
 <script src="build.js"></script>
 
 </head>
--- a/browser/index.html
+++ b/browser/index.html
@@ -16,6 +16,7 @@
 <script src="https://cdn.rawgit.com/CamHenlin/TouchPolyfill/master/touchpolyfill.min.js#sabaki-v0.11.0"></script>
 <script src="../node_modules/sigma/build/sigma.min.js"></script>
 <script src="../node_modules/sigma/build/plugins/sigma.renderers.customShapes.min.js"></script>
+<script src="../node_modules/clipboard/dist/clipboard.min.js"></script>
 <script src="build.js"></script>
 
 </head>
--- a/browser/index.js
+++ b/browser/index.js
@@ -915,13 +915,10 @@ sabaki.prepareCleanMarkup = function() {
 }
 
 sabaki.prepareClipboard = function() {
-    document.addEventListener("copy", function(e) {
-      e.clipboardData.setData("text/plain", sabaki.saveFileToSgf());
-      e.preventDefault();
-    });
-    document.addEventListener("paste", function(e) {
-      sabaki.loadFileFromSgf(e.clipboardData.getData("text/plain"), false, true);
-      e.preventDefault();
+    new Clipboard(".saveClip", {
+        text: function(trigger) {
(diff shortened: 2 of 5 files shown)


C0DEHERO:
@yishn could you take a look at this, and tell me what the preferred implementation would be?

yishn:
Hello @C0DEHERO, thank you for your work! I thought this was WIP, so I didn't comment. 

There are certain drawbacks when hacking into the browser 'copy' and 'paste' events like this. For example, when you're writing a comment, you can't copy and paste into the comment box anymore. The browser will intercept the actions and try to copy/load the SGF. This is not good. This applies to all text boxes.

I thought of using [clipboard.js](https://github.com/zenorocha/clipboard.js) for copying the SGF to the clipboard. As for loading SGF from the clipboard, I see no other alternative than to provide a text box and let the user paste the contents into it.  The most simple way (but not very user-friendly) is to use a `prompt()` box, but it will work.

C0DEHERO:
Oh, right. I totally forgot about comment copy/pasting. It would of course be possible to check if the clipboard content is a valid SGF and then decide whether to call the loading function or to just pass the event on. I wasn't sure if it would be okay to use clipboard.js, but since it is, I'll use that instead.

> I thought this was WIP, so I didn't comment.

My bad. I should've explicitly asked for feedback in the original comment. I figured it'd be better to ask rather than finish a full implementation that you wouldn't like.

yishn:
No sweat, it's a good thing you asked!

C0DEHERO:
So, this code works, but I'm not sure it's the nicest. Clipboard.js wants to its own event handler, so I had to add a class to the 'save to clipboard' button in menu.js unless I wanted to query `ul.popupmenu.show li` and then check for the `textContent` of the selected button, which would have been a lot uglier. So this is probably the better implementation. I also don't like the `click: () => ""`, but I can't see how to remove the need for an empty callback function without modifying menu.js more than I probably should for just one button.

browser/index.html
```diff
@@ -16,6 +16,7 @@
 <script src="https://cdn.rawgit.com/CamHenlin/TouchPolyfill/master/touchpolyfill.min.js#sabaki-v0.11.0"></script>
 <script src="../node_modules/sigma/build/sigma.min.js"></script>
 <script src="../node_modules/sigma/build/plugins/sigma.renderers.customShapes.min.js"></script>
+<script src="../node_modules/clipboard/dist/clipboard.min.js"></script>
```
yishn: Instead of including clipboard.js with a `<script>` tag, I'd rather have it required as a node.js module in `index.js`:

~~~js
const Clipboard = require('clipboard')
~~~

browser/index.js
```diff
@@ -914,6 +914,14 @@ sabaki.prepareCleanMarkup = function() {
     })
 }
 
+sabaki.prepareClipboard = function() {
+    new Clipboard(".saveClip", {
```
yishn: Please use single quotes to match with the existing coding conventions.

browser/view.js
```diff
@@ -1110,6 +1110,15 @@ exports.openHeaderMenu = function() {
             label: 'Download SGF',
             click: () => sabaki.saveFile()
         },
+        {
+            label: 'Save to Clipboard',
+            type: 'saveClip',
+            click: () => ""
```
yishn: Please use `() => {}` for an empty closure.

browser/view.js
```diff
@@ -1110,6 +1110,15 @@ exports.openHeaderMenu = function() {
...
+        },
+        {
+            label: 'Load from Clipboard',
+            click: () => sabaki.loadFileFromSgf(window.prompt("Paste the sgf here"))
```
yishn: Single quotes here as well

PR #171, opened 2017-03-04
Call out good and bad moves on board

dfannius:
I have an SGF file with a lot of joseki variations I am learning. Sabaki's "ghost stones" are great for this, as I can immediately see all the interesting moves on the board instead of having to descend into each branch. But many of my branches are bad moves along with reasons that they are bad and ways to punish them. I would like to be able to distinguish bad moves from normal moves at a glance when looking at the ghost stones.

This commit causes the ghost stones for bad (BM) and good (TE) moves to be tinted slightly red and green respectively, so you can identify them easily without descending into each branch. It already has made it much easier for me to use my joseki SGF file.

I know very little of both JavaScript or CSS, so I have no illusions that this is good code, and I'm submitting it mostly as a proof of concept. It's your call whether to use the feature, change it somewhat, put it behind a preference, etc. If you reject it I don't mind just keeping it around in my local build.


yishn:
Hey, that's pretty neat! We can also add a menu item to toggle this on and off. The colors also reflect the move color which is pretty cool, but what about the sibling moves? Shouldn't they get the tint as well? 

Also, I would also like to color 'interesting' and 'doubtful' moves. (Maybe yellow for interesting moves and purple for doubtful moves?) Do you want to continue working on this or should I take over?

dfannius:
Originally I intended to also colorize sibling moves, but then was afraid that it might make things too busy. What I really want to know is "which of the displayed moves I could make now are bad", rather than "which of the moves my opponent could have made instead of the last one are bad".  Displaying both at the same time seemed potentially confusing. But I could try it out.

I'm happy to keep working on it. How about yellow for doubtful (going with the traffic light analogy) and blue for interesting (I think purple might look too close to red)? At some point we'll run out of easily distinguishable colors, especially given translucency, but maybe four is still okay.

yishn:
I say we can try it out! Yeah, yellow and blue colors do make sense.

dfannius:
I added colorization for doubtful and interesting moves and I also colorized siblings. In my joseki SGF file I'm now using the interesting tag for moves that depend on ladders or other outside context, which works well.

If I try setting up a node with lots of child and sibling moves of all types, things get really busy and it's kind of hard to tell what's going on. But any single node is unlikely to have many different sorts of annotations in practice. Your call whether you want to keep sibling colorization.

yishn:
I've cleaned up some code and changed the colors. Thank you very much!

dfannius:
Great! I just did a pull with your changes and noticed a couple of things:

- Ghost colors for White and Black stones are now the same rather than Black ones being darker. I find this a little confusing. Is it intentional?
- Sibling ghost nodes don't seem to be colorized.

yishn:
Yeah, I played with it for a while and I decided to leave sibling colorizations out. I'm still playing with the colors though. I want them to be both good looking and distinguishable.

Issue #172, opened 2017-03-05
Request: pop up comment menu by right clicking on last-played stone


dfannius:
Now that move colorization has made it useful to mark moves with the "Good Move" et al annotations, it would be nice to have a more elegant way to set those annotations. In the current UI, I make the move, click on the edit icon, click on the down arrow, click on Good Move, then click on the X of the bottom bar to exit edit mode, which is 5 actions (or worse, I forget the last one, and then start placing stones instead of making moves, and have to undo all of that). It would be more convenient if right-clicking on a just-played stone would bring up the same menu, which would bring us down to 2 actions.

I implemented this myself locally and I think it works great from a UI viewpoint, but I don't really know JavaScript so the code is very hacky and not worth submitting as a pull request. So instead I am submitting it as an issue :) All I really did was add a position parameter to `openCommentMenu`, update the way it is currently called, and add another call to it at the bottom of `vertexClicked`, analogous to the way the Remove Node code works.

yishn:
Your implementation sounds good. :wink: You can totally send in that PR.

PR #173, opened 2017-03-05
Pop up comment menu by right clicking on last-played stone

dfannius:
Implements the request in issue #172.


yishn:
LGTM, this feature is something I never I missed!

PR #177, opened 2017-03-15
Parse GIB / NGF files

fohristiwhirl:
This commit allows the parsing of GIB files from Tygem. The file format is poorly documented so everything is somewhat crude.

I am by no means a great JS coder, in fact I started learning the language a month ago. If you wish to disregard this, I would have the greatest of sympathy. But I put this here in case you're interested.

Edit: now updated with an NGF parser as well for wBaduk.


modules/gib.js
```diff
@@ -0,0 +1,217 @@
...
+
+            if (handicap >= 2 && handicap <= 9) {
+                root.HA = [handicap.toString()]
+                root.AB = []
+
+                let points = handicap_points(19, handicap, true)
+
+                for (let i in points) {
+                    let x = points[i][0]
+                    let y = points[i][1]
```
yishn: You can use destructuring like so:

~~~js
let [x, y] = points[i]
~~~

modules/gib.js
```diff
@@ -0,0 +1,217 @@
...
+                    root.RE = ['B+']
+                }
+            }
+
+        } else if (line.slice(0, 3) === 'INI') {
+
+            let setup = line.split(' ')
+            let handicap = 0
+            try {
+                let p = parseInt(setup[3])
```
yishn: Avoid using `parseInt()` without radix; it can lead to unexpected results such as `parseInt('080') == 0` in older browsers. [More information here.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt) Just use `parseFloat()` instead (and `Math.round()` if necessary).

modules/gib.js
```diff
@@ -0,0 +1,217 @@
...
+        } else if (line.startsWith('\\[GAMERESULT=')) {
+
+            let score = null
+            let strings = line.split(' ')
+
+            // Try to find score by assuming any float found is the score.
+
+            for (let i in strings) {
+                try {
+                    let p = parseFloat(strings[i])  // Does this ever throw?
```
yishn: No, both `parseFloat` and `parseInt` do not throw in Javascript, so the `try` block is unnecessary. But they may return `NaN`.

modules/gib.js
```diff
@@ -0,0 +1,217 @@
...
+            }
+            catch (e) {}
+
+            if (handicap >= 2 && handicap <= 9) {
+                root.HA = [handicap.toString()]
+                root.AB = []
+
+                let points = handicap_points(19, handicap, true)
+
+                for (let i in points) {
```
yishn: Use `of` keyword to loop over the elements in the array:

~~~js
for (let point of points)
~~~

fohristiwhirl:
I believe that's everything.

This is based on something I wrote a while ago in Python:

https://github.com/fohristiwhirl/xyz2sgf

If you're interested, it also has rudimentary parsers for NGF and UGF (although the latter format can have variations I think, which I don't parse). I believe NGF is simple though and could submit an additional pull request for that, if you like.

fohristiwhirl:
Oh I should mention: I don't fully understand the Sabaki internal representation so I haven't touched tree.collapsed, tree.current, or tree.parent. It seems to work as is.

yishn:
The other properties of `gametree`, like `collapsed`, `parent`, and `current`, only concerns with how Sabaki deals with game variations (subtrees). It seems like `gib` and `ngf` files don't support variations? (Is that right?) So they don't matter.

fohristiwhirl:
I'm not 100% sure if GIB and NGF ever have variations, I've never seen them. They are rather poorly documented file formats.

yishn:
I really like the idea that 'file format modules' should have a uniform structure for parsing/loading, so one can simply replace, say module `sgf` with module `gib` without problems. Maybe these functions should be available in a file format module:

~~~js
xyz.meta = {name: 'Blah', extensions: ['xyz']}
xyz.parse = (content, callback = () => {}) => gametree
xyz.parseFile = (filename, callback = () => {}) => gametree
~~~

where `callback` is called at regular intervals with a float value between 0 and 1 to indicate how far the parsing process is progressing. We can merge the `meta` information of all file format modules to build a universal extensions filter.

I will update module `sgf` accordingly. (Right now, the `sgf` parser is a two step process: Tokenize, and then parsing.)

modules/gib.js
```diff
@@ -0,0 +1,145 @@
...
+        } else if (line.startsWith('\\[GAMEWHITENAME=') && line.endsWith('\\]')) {
+
+            let s = line.slice(16, -2)
+            root.PW = [s]
+
+        } else if (line.startsWith('\\[GAMECONDITION=')) {
+
+            // Hard-coded the common komi cases.
+            // For better results, we could do a regex here instead.
+
```
yishn: Yeah, this does seem a little restrictive... Since it seems like there's an obvious pattern to it. 😉

fohristiwhirl:
OK, changed. It should be noted that GIB files come in many languages, I'm only parsing komi for the English case, otherwise we'll just leave it unspecified.

fohristiwhirl:
On current assumptions, both GIB and NGF loading should be so quick as to not require a progress bar.

yishn:
I guess you're right, but I still want this piece of code

~~~js
trees = xyz.parse(content, progress => {
    if (progress - lastprogress < 0.1) return

    view.setProgressIndicator(progress, win)
    lastprogress = progress
}, ignoreEncoding ? null : undefined).subtrees
~~~

to work for all `xyz` in `[sgf, gib, ngf]`. `gib` and `ngf` can ignore the extra arguments.

fohristiwhirl:
OK I changed the function signatures. I wasn't sure if you wanted a third argument to be `encoding` or `ignoreEncoding`, so I left it for now.

fohristiwhirl:
I've looked at as many files as I can and I think I've discovered a couple of things about how GIB files store their data, I may be able to improve handling of non-English files...

yishn:
Cool, great work! Do you have some sample gib and ngf files that I can use to add some tests?

fohristiwhirl:
I've added some files to the test directory.

wBaduk works well enough that NGF files are easy to get. But the Western Tygem client seems to have its save button broken. However, one can find some games on the net.

One thing I've not dealt with really is character encoding, i.e. if a Chinese game is in GB2312, the names will be garbled (but it will otherwise work). I don't believe either GIB or NGF directly specifies its encoding, though it's possible all GIB games marked as "China" are supposed to be GB2312, for instance. But one can't really rely on that?

I think loading with the possibility of garbled names is better than not loading at all?

fohristiwhirl:
If you like, I could investigate whether some npm package like `detect-character-encoding` would be helpful.

yishn:
I really want to make it work. Can you use [jschardet](https://github.com/aadsm/jschardet) to see if it can correctly detect the Chinese/Korean files you got?

fohristiwhirl:
I'll give it a go.

fohristiwhirl:
This seems to work. I've not touched the package dependency list, so it would need updating.

fohristiwhirl:
Sorry for all the commits, I think I'm about finished. There's a lot of redundancy in the file so I now check 2 different lines for result / komi, in case one such line is missing (which I've seen).

fohristiwhirl:
I was able to improve charset detection for NGF by only sending the first few lines to be checked. (The rest of the file is in ASCII.)

GIB seems to work as is, probably because it's mostly numbers.

yishn:
It doesn't seem like jschardet is doing so well on the Korean `gib` file. But oh well, it seems like parsing `gib` and `ngf` files isn't an exact science either, so whatever...

PR #180, opened 2017-03-17
Use boardsize when creating Board for .getHandicapPlacement()

fohristiwhirl:
This fixes handicap placement for NGF when the board is not 19x19. (I
think GIB is always 19x19).


yishn:
Thanks!

PR #181, opened 2017-03-20
Preact Rewrite

### LINKED ISSUE #79: Custom Labels

### LINKED ISSUE #112: API for the web version of Sabaki?
Is there documentation for the API to the web version of Sabaki? I'd love to use it in some projects I am working on, but need the ability to manipulate the board directly. 

i.e.
`sabakiGame.playAt(1, 2);`
`sabakiGame.undo();`

### LINKED ISSUE #140: Allow both sides to be played by engines at once
Currently if you assign an engine to one colour, the other colour will have its engine setting cleared. It should be possible to attach two engines (or two copies of the same engine) to the same Sabaki game and have them play against one another, rather than always requiring a human player.

### LINKED ISSUE #178: Sabaki hangs when adding a non-valid game to a collection
This is not directly related to my current pull request... i.e. this problem exists whether that request is accepted or not.

When in "Manage Games", if we add an existing file, and the file is not valid SGF, Sabaki hangs.

The problem is somewhere around:

https://github.com/yishn/Sabaki/blob/master/browser/view.js#L1416

### LINKED ISSUE #185: Initial board saves its size like SZ[19:19]
If you open Sabaki, maybe play a few moves, and save, the file will have its size saved in the long format, e.g. SZ[19:19]. This doesn't happen if you ever select "New".

For compatibility with older SGF editors, it would be better not to save the long format where possible.

Update: I think I can fix this actually... it's very minor changes to sabaki.getEmptyGameTree:

    let sizeProperty = size[0] + (size.length > 1 ? ':' + size.slice(-1)[0] : '')

And then use that inside the template...

    `SZ[${sizeProperty}]`,

If you agree, I can make a pull request if you wish, but it's a simple enough change we could avoid it...

### LINKED ISSUE #194: clear_board and boardsize are sent in wrong order over GTP
According to the GTP specs, after "boardsize" the contents of an engine's board are arbitrary. So the correct order should be:

boardsize 19
clear_board

But actually Sabaki does it the other way around.

This can be fixed by swapping lines 1059 and 1060 of index.js.

### LINKED ISSUE #197: AB tags don't accumulate
I just opened the SGF file http://lifein19x19.com/forum/download/file.php?id=2310 in Sabaki 0.30.0, and many Black stones are missing from the initial position (you can see the file displayed in Eidogo at http://lifein19x19.com/forum/viewtopic.php?p=119544#p119544).

Studying the SGF file, my guess is that the second AB tag (containing only `[ic]`) is blowing away the first AB tag (containing `[pp][pn][qq][pj][pd][fc][db]`) instead of adding to it. I looked briefly at the SGF spec but couldn't tell whether AB tags are supposed to accumulate or replace earlier ones. In any case it seems that at least one other program (the file says it was created with MultiGo) assumes that they accumulate. I believe that I have opened the file in CGoban before and it loaded as intended.

Presumably this is the case with AW tags as well.

yishn:
This PR tracks progress on an experimental [preact](http://preactjs.com/) rewrite of Sabaki's UI. You can try it out by checking out this branch and `npm install`.

### Todo
- [x] Updater should only notify if latest version is higher than local version
- [x] Investigate bug in `GtpConsole` where `<select/>` is not wide enough after swapping engine
- [x] Generate move when user plays a move
- [x] Generate move when closing `InfoDrawer`
- [x] Resize window when showing/hiding sidebars
- [x] Save window dimensions
- [x] Better instructions on how to stop engines from playing
- [ ] Rethink attach engines menu
- [x] Release Sabaki v0.30.0 beta

### Web

- [x] Reimplementation of `dialog` module
- [x] `Menu` shim
- [x] Dummy `jschardet` and `iconv-lite` modules
- [x] Load and save files
- [x] `clipboard` shim
- [x] `shell` shim
- [x] `setting` should work with local storage
- [x] Keyboard shortcuts

### Documentation
Create basic documentation files for:

- [x] `sabaki`
- [x] `gametree`
- [x] `Board`
- [x] `boardmatcher`
- [x] `deadstones`
- [x] `dialog`
- [x] `fileformats`
- [x] `helper`
- [x] `setting`
- [x] `sound`

Fixes #178, fixes #185, fixes #194, closes #79, closes #140, closes #112, closes #197.


alreadydone:
As for #140, it's still not possible to set up an engine to play both sides automatically? That would speed things up because of tree reuse etc. for Leela Zero, for example.

yishn:
Please open a new issue about this.

PR #183, opened 2017-03-22
Allow stone overwrites

### LINKED ISSUE #182: Should allow moves to overwrite previous contents.
(This is a tiny issue but also easy to fix, I think.)

According to red-bean... http://www.red-bean.com/sgf/go.html

`When a B (resp. W) property is encountered, a stone of that color is placed on the given position (no matter what was there before).`

Currently Sabaki ignores moves that overwrite a stone that is already there. I believe this can be fixed to match the recommended behaviour with 2 tiny changes:

* Commenting out line 382 in board.js (which checks for a stone already present).
* Changing the word `this` on line 395 to be `move` instead (so that the new board is used to determine which chains to remove, if capturing).

This test file works after both these changes are made:

```
(;GM[1]FF[4]CA[UTF-8]SZ[19]
;B[jg];W[kh];B[jh];W[ki];B[ji];W[jj];B[kg];W[lg];B[jf];W[kf]
;B[je];W[ke];B[cn];W[jd];B[ci];W[ie];B[cf];W[if];B[ck];W[ig]
;B[en];W[jg])
```

fohristiwhirl:
This allows stones to be overwritten.

I trust that makeMove() is never called as a legality checker, i.e. you never call it just to check the result isn't null. Currently, that seems to be true.

Closes #182.

### DIFF
--- a/modules/board.js
+++ b/modules/board.js
@@ -380,9 +380,6 @@ class Board {
 
         if (sign === 0 || !this.hasVertex(vertex)) return move
 
-        // Next line commented out to allow moves to overwrite stones...
-        // if (this.get(vertex) !== 0) return null
-
         sign = sign > 0 ? 1 : -1
         move.set(vertex, sign)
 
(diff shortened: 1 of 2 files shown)


yishn:
Thanks! I don't think I do legality checks with makeMove (It would just tell me whether a stone occupied the vertex anyways, which is not really helpful). Just one thing: Can you just remove the commented out lines? It won't help anybody to let it stay in the code base.

PR #236, opened 2017-05-21
Fix how parseCommand sends arguments to Command class

### LINKED ISSUE #235: GTP console replaces spaces with commas
I attached the Leela GTP engine to my client. When I try to adjust the time settings with the "time_settings 20 30 1" command in the GTP console, Sabaki automatically changes my command to "time_settings 20,30,1". This is the wrong syntax for Leela and results in a syntax error.

fohristiwhirl:
Should fix #235


yishn:
Thank you!

PR #273, opened 2017-09-09
trying to fix incorrect komi value sent to gtp

geovens:
When both changing the komi value of an opened sgf file and attaching a gtp engine _during a single visit of the info drawer_, the old komi value remaining in inferredState is used in syncEngines() after the button 'OK' is clicked. This causes the old komi value being sent to gtp in the initialization of the engine. I try to fix this bug with this commit. Not sure my fix is proper though. Please check, thanks.

### DIFF
--- a/components/App.js
+++ b/components/App.js
@@ -1885,7 +1885,7 @@ class App extends Component {
     attachEngines(...engines) {
         let {engineCommands, attachedEngines} = this.state
 
-        if (helper.vertexEquals([engines[1], engines[0]], attachedEngines)) {
+        if (helper.vertexEquals([...engines].reverse(), attachedEngines)) {
             // Just swap engines
 
             this.attachedEngineControllers.reverse()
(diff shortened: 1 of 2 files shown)


yishn:
Nice catch! `inferredState` has shown to be a terrible idea on my part over and over again. But I don't think your fix is gonna work since `root` is not even defined here. You can use `this.state.treePosition[0]` instead.

yishn:
That said, I notice that komi won't get updated when changing it in the info drawer if there are already engines attached. I remember vaguely that this bug has been discovered before and I fixed it, but it probably got lost in the rewrite. I'll open a separate issue about that.

geovens:
Great, if that could be fixed, it would make game analysising with Leela more easy for me. Should I close this pull request then? I want to create another one for theme.
Maybe unrelated: when I type any commend (like "komi 3.5") in the GTP console, it replies "blocked command". Is this a reply by Leela or Sabaki? Do you know why I can't do it myself?

geovens:
or could I just change "let komi = gametree.getRootProperty(root, 'KM')" to "let komi = gametree.getRootProperty(this.state.treePosition[0], 'KM')" ?

yishn:
> or could I just change "let komi = gametree.getRootProperty(root, 'KM')" to "let komi = gametree.getRootProperty(this.state.treePosition[0], 'KM')" ?

Yes, please do that.

> Maybe unrelated: when I type any commend (like "komi 3.5") in the GTP console, it replies "blocked command". Is this a reply by Leela or Sabaki? Do you know why I can't do it myself?

The gist is that you should change the komi settings from within Sabaki. But it doesn't work currently, hence the issue I opened at #275.

yishn:
Hmm... it doesn't seem like this does the trick, though... The komi is still not updating when changing komi and engine at once in the info drawer. I'm gonna investigate some more and merge this into #275. Please subscribe there for updates.

yishn:
Apparently, this is indeed working, my bad!

Issue #274, opened 2017-09-09
Allow more than 10 seconds per move in Autoplay.


emauton:
I am on v.0.31.2 on Ubuntu Trusty.

Right now autoplay seems to max out at 10s, per
https://github.com/yishn/Sabaki/blob/dd89afa42beb222bd20391aac9501eade76939de/components/bars/AutoplayBar.js#L25
  https://github.com/yishn/Sabaki/blob/cd51033f6e44942b13aa5d9837300a8aa44f2e93/components/bars/AutoplayBar.js

I'd prefer to be able to use e.g. 30 or 60s - that gives me more time to try to read and understand what's going on when watching through pro game SGFs.

Any technical reason for 10s, or can I take on to patch that? What's the thinking behind the formula used there (I can add a comment to explain that too)? Thanks!

Cian

yishn:
Probably because I think the input box is too small for longer values (i.e. three digits). 10s seems rather restrictive in hindsight. Do you want to change that? We can take it up to, say 99s, and prevent the user from entering decimal numbers. You also need to change the corresponding attributes on the input box itself: 

https://github.com/yishn/Sabaki/blob/cd51033f6e44942b13aa5d9837300a8aa44f2e93/components/bars/AutoplayBar.js#L94-L95

emauton:
Sounds good. I'll have a shot at that sometime over the weekend. Thanks!

PR #276, opened 2017-09-09
Allow up to 99s per move in Autoplay, step by 1s

emauton:
Fixes https://github.com/yishn/Sabaki/issues/274


yishn:
Thank you very much!

PR #278, opened 2017-09-10
add cartoon theme

geovens:
By the way, how about adding a link to the theme directory page in the app?


yishn:
Good idea! I'll open an issue.

PR #291, opened 2017-10-07
change color of special inactive nodes

geovens:
Use darkened blue/purple color to inactive nodes that have comment/bookmark on it. This could help in showing where all the comments are, when reading others' game reviews.


yishn:
Nice addition!

PR #298, opened 2017-10-22
[#296] Update guides - add Ray engine

### LINKED ISSUE #296: [Docs] Add Ray to /docs/guides/engines.md
https://github.com/zakki/Ray
This bot play about the same level as Leela and AQ and works fine in Sabaki, so may be it can be added to 
https://github.com/yishn/Sabaki/blob/master/docs/guides/engines.md ?
Example of config line:
--const-time 60 --thread 4 --reuse-subtree --tree-size 262144

aTan-aka-Xellos:
Update for issue https://github.com/yishn/Sabaki/issues/296

Closes #296


yishn:
Thank you! I think "(for optional parameters - check engine documentation)" applies to all engines, so maybe we can add this line as an extra paragraph on top?

aTan-aka-Xellos:
@yishn updated

PR #316, opened 2017-12-29
Engine synchronization

yishn:
Currently, to update the attached engine to current board arrangement, Sabaki will recreate the board by placing stones from upper left to lower right. This causes problems in some engines which use neural networks (see #294).

This pull request introduces a new module `enginesyncer` which will be responsible for engine synchronization. It will try to update the board arrangement in the following order:

1. Try to update engine board with one move
2. Try to recreate board by replaying moves and additional stones in order of the game tree
3. If all above fails, try to recreate board by placing stones from upper left to lower right

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 
 * Add engine integration protocol
 
+**Changed**
+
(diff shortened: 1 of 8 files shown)


killerducky:
Oh I was just starting to look into this, but it seems you've already got a solution, great work!

PR #339, opened 2018-02-17
Use Q4 as 3rd handicap stone normally

### LINKED ISSUE #337: Placement of handi stones
Is there a way to change where the handicap stones are, specifically for  3 handi games? Crazy Stone has the top left corner empty, while Sabaki has the bottom right corner empty. (I also think KGS does it the same way as Crazy Stone.)
The issue came up when I wanted to play Crazy Stone against other engines in Sabaki and the placements were different.

fohristiwhirl:
Closes #337

### DIFF
--- a/src/modules/board.js
+++ b/src/modules/board.js
@@ -418,7 +418,8 @@ class Board {
         return move
     }
 
-    getHandicapPlacement(count) {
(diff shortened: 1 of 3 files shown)


yishn:
Looks good, thank you very much!

PR #362, opened 2018-04-03
Add an option to toggle the hardware acceleration

ohyou:
**Goal:**

To let users turn hardware acceleration on/off. By default, it's turned on, but turning it off may help people with slower PCs, and it also lets capturing software (e.g. OBS) to capture the Sabaki window.

**Description:**

This PR adds an `app.enable_hardware_acceleration` key to the `setting`, which controls whether the app will be launched with hardware acceleration enabled. To control this setting, a checkbox was added to the PreferencesDrawer. The default state is left unchanged.


yishn:
Thank you for the PR!

PR #376, opened 2018-04-24
get ponder variations from leelasabaki

5525345551:
At the moment Sabaki only shows the variations for Leela.

This will allow ponder variations to update the gametree.

Also:
Add labels A,B,C, ... in the root gametree for variations if specified in leelaSabaki
Add option to exit without asking to save. I personally find that really annoying :)

Needs the commits in my branch of LeelaSabaki to work.

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -2047,6 +2047,20 @@ class App extends Component {
             }
         }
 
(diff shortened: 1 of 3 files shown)


hadim:
Any idea if this will be merged?

PR #377, opened 2018-04-24
Rotation functions

fohristiwhirl:
Rotation functions


fohristiwhirl:
Meanwhile, an extremely naive method for the App component looks like this:

```javascript
    rotateClockwise() {

        const rotation = require('../modules/rotation')
        
        let root = gametree.getRoot(this.state.treePosition[0])
        let trees = gametree.getTreesRecursive(root)

        for (let tree of trees) {
            for (let node of tree.nodes) {
                rotation.rotateNodeClockwise(node, 19)      // FIXME - get actual size; check if height == width
                node.board = null
            }
        }
    }
```

This does basically work.

fohristiwhirl:
Should I add my (naive) method and menu item to the PR?

fohristiwhirl:
So I suggest we simply disallow rotation while engines are attached. I'm not sure how to check for that though.

yishn:
Will it be a problem if engines are attached? Theoretically, the board state will be synced via `enginesyncer`, the worst case being the whole game being replayed to the engine.

yishn:
I can tell you that it won't be a problem theoretically, so we don't have to disallow rotation when engines are attached. But I need to test it, of course.

fohristiwhirl:
I tested it with a random move generator which makes legal moves or passes when there aren't any - everything worked as expected.

fohristiwhirl:
OK I got around to implementing anticlockwise properly.

yishn:
Thank you, that's much more than you promised!

Issue #371, opened 2018-04-18
Detach seems broken


fohristiwhirl:
I upgraded to current (i.e. the tip of the master branch). But detach seems broken. Not sure if it was broken before.

Reproduction

* Start a game with 2 engines.
* Let them play a few moves.
* Press escape.
* Detach.
* Play a move by yourself.

The engines will now start again.

yishn:
Thanks for the report. I think it has something to do with the recent restructuring of engine handling in master.

5525345551:
Phew,

I thought it was something I broke when i started adding hacking on Sabaki and leelasabaki,
What a relief to know it was broken before. :)

PR #378, opened 2018-04-25
fix#371 detach engine issue

5525345551:
Also send command 'quit' to the engine so it exits the process.


yishn:
Thank you for your pull request, however issue #371 is actually caused by a bug in the attachEngines() function. It should detach attached engines if passed null as argument. Furthermore, there's no need to send the quit command manually since the gtp module will do that automatically when you stop the engine.

PR #389, opened 2018-05-15
Make things more robust when size > 26

fohristiwhirl:
(For the rotation functions.)

When boardsize is > 26, we can't simply use ASCII values for min() and max() because SGF uses lowercase values to mean lower numbers, but ASCII has them as higher numbers.

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -466,7 +466,7 @@ class App extends Component {
 
         let {extname} = require('path')
         let extension = extname(filename).slice(1)
-        
+
         let gameTrees = []
         let success = true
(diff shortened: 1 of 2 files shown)


fohristiwhirl:
I also noticed [this line](https://github.com/fohristiwhirl/Sabaki/blob/994f0ef3548b08b22c996330eb98dce8b56d2559/src/components/App.js#L1450) stops setGameInfo() from accepting sizes > 25, so I removed an unneeded call to it from App.js.

yishn:
Thanks!

Issue #336, opened 2018-02-12
Allow sorting games by number of moves in the game (and/or by file size)


wctgit:
This may be a little specific, but I happened to be looking for a way to find LeelaZero self-play games which ended with a ladder early in the game. I thought by using Sabaki's 'Manage Games' feature, I might be able to find some games like that by looking at the board previews. Unfortunately, that doesn't show the end position as I thought it did.

So, the next idea I had was to sort the games by number of moves, since early ladders lead to short games. I got all excited when I saw that Sabaki does indeed include a Sort feature. But it doesn't include number of moves... ah, so close! Anyway, I thought I'd submit an issue for this since it was a real-world use case.

As a kind of cheap work-around, even sorting by the size of the file would probably help. I'm going to try that on my file system to see if anything turns up, in the meantime.

gcentauri:
Hi there!  I'm a fan of Sabaki as well, and thought I'd try my hand at contributing to the project.  I have submitted PR #391 to sort games by number of moves.

PR #391, opened 2018-05-17
add sort by number of moves

gcentauri:
addresses issue #336 

the logic for the sorting menu could probably be refactored to be a bit cleaner, but adding the 'moves' check at the bottom seems to work.  happy to continue working on it if there's more to improve here!


gcentauri:
i'd be happy to try writing some tests around this module as well, but i'm unfamiliar with the test framework.  i'll do a little bit of digging

src/components/drawers/GameChooserDrawer.js
```diff
@@ -297,6 +298,10 @@ class GameChooserDrawer extends Component {
                             s = x1 < x2 ? -1 : +(x1 !== x2)
                         }
 
+                        if (property === 'moves') {
```
yishn: We can include this `if` clause as an `else if` between line 296 and 297.

yishn:
Thank you for your contribution! It looks good, I just left some minor comments. 

As for testing, currently we have no framework in place for testing components. One solution would be to extract the logic for game sorting into a separate js file in `src/modules` and write tests for it. But if this is too much work, we can forgo it.

yishn:
That's great, thank you very much!

> Some of the test game data available didn't seem to work with the SGF property scheme the selector is using.

What do you mean by that exactly?

> I'm wondering if it wouldn't be worth creating some kind of GameCollection class to hold games and implement the sort method on it? Maybe that is something you already have in mind.

Sabaki uses plain objects and arrays to encode SGF data, and I'd like to keep it like that.

gcentauri:
> What do you mean by that exactly?

Well for example, the file euc-kr.gib created a game tree where the value of the BR property was 1±Þ.  Perhaps this has some meaning I don't understand, but it looked like the file indicated that the "Level" of the player was "17".  Regardless, that string wouldn't sort properly with the function there currently.  Perhaps I am missing a step in parsing files that aren't SGF?

> Sabaki uses plain objects and arrays to encode SGF data, and I'd like to keep it like that.

After thinking about it more I agree.  The important part is just the ability to compare, there's not really any need to abstract something else.

fohristiwhirl:
> Well for example, the file euc-kr.gib created a game tree where the value of the BR property was 1±Þ

That's a failure to detect the proper encoding in the GIB. But in any case, properties like BR can in reality be completely arbitrary.

yishn:
Maybe let's only use the SGF files for testing for now. I'd say the other formats don't have priority.

gcentauri:
Okay, if you'd like some more robust tests, I can put together a small collection of SGF files to include in the repo.  As far as I know, game records are free to use data, correct?  Barring any commentary I suppose.  I have a collection of my own from KGS, and a large collection of Pro games.  Let me know if there's any preferred source for the actual data files.

yishn:
That's a good idea, we probably just need a small subset of games for testing.

gcentauri:
I created a few sgf files with Sabaki and changed the tests to sort them.  Hope it all looks okay!  Thanks for making Sabaki!

yishn:
Thank you very much! I probably thought there would be no need to sort them with natsort. Natsort ensures that numbers are sorted correctly, e.g. that 'Game 10' comes after 'Game 2'. Since it's unlikely that player names contains numbers, I didn't sort them with natsort. But for consistency's sake, maybe we should... What do you think?

gcentauri:
User names from go servers could certainly contain numbers, though the sequence is likely meaningless...  Shodan, Shodan1, Shodan01, Shodan2, Shodan10... I guess it isn't probably something a user would notice too much?

gcentauri:
I am doing some more refactoring on the sort module, but I missed the trick for stable sort.  I will try to finish that tonight and push up another commit.  I went ahead and grouped all Name type properties to use natsort

gcentauri:
Hope I didn't go too overboard refactoring... let me know what you think! I can squash all of this into a single commit if you prefer to merge that way.

yishn:
I like the new API of `gamesort`, thank you again for your hard work!

PR #419, opened 2018-07-04
Retain SZ property when flattening (fixes #418)

### LINKED ISSUE #418: Flatten discards SZ property
The SZ property (board size) should be retained when flattening.

dfannius:


yishn:
Thank you!

PR #426, opened 2018-07-18
Correct "Respository" misspelling

dpflug:
Trivial thing, but it was bugging me.


yishn:
😅 Thank you so much for correcting!

PR #451, opened 2018-10-07
Analysis

yishn:
Analysis

- Performance improvements to the GTP console
- Press <kbd>F4</kbd> to display ponder information using the `lz-analyze` GTP command
- Add variation to game tree by right click > 'Add Variation'
- Add a setting to let user to turn off automatic engine move generation
- Remove support for LeelaSabaki

Only the very latest version of Leela Zero's next branch is being supported, so hopefully Leela Zero v0.16 is coming out soon.


hadim:
I have been testing it for a few days now and it's a great feature!

Sometimes the heat map doesn't appear after pressing F4... I couldn't find a reliable way to reproduce it consistently. I'll keep you in touch if I notice something.

hadim:
It looks all normal in the console. The bug does not happen often.

PR #452, opened 2018-10-11
Add gamepads support using the HTML5 Gamepad specification.

hadim:
I have tested it with a PS4 controller on Chrome and it works well.

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -32,6 +32,7 @@ const helper = require('../modules/helper')
 const rotation = require('../modules/rotation')
 const setting = remote.require('./setting')
 const sound = require('../modules/sound')
+const { GobanGamepad } = require('../modules/gobanGamepad')
 
 class App extends Component {
     constructor() {
@@ -109,7 +110,14 @@ class App extends Component {
             // Info Overlay
 
             infoOverlayText: '',
-            showInfoOverlay: false
+            showInfoOverlay: false,
+
+            // Cursor
+            cursor: {
+                vertex: null,
+                marker: null,
+                previousCursor: null
+            }
         }
 
         this.events = new EventEmitter()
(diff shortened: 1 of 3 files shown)


hadim:
Mapping for other controllers might need to be adjusted. It's pretty easy to do if needed.

yishn:
Hi, I was on vacation, I've just roughly gone through the changes. You're using the cross tool to draw the 'cursor' which changes the game tree, i.e. saving the file while the cursor is visible will save the cross on a vertex as well. We need some sort of cursor state on a vertex first, distinguishable from SGF markers.

Also, maybe the big picture idea is to make the board keyboard-accessible first, then try to make it work with gamepads?

hadim:
Yes, I use cross tool to have a proof of concept quickly but I agree is not optimal. In that case, I think modifications should be done in Shudan right? Where do you think would be best to add such a feature?

As for the keyboard-accessible, you're right at it's also on the TODO list. The thing is that the gamepad API is pretty independent of any keyboard stuff. So I was thinking of having a `src/modules/gobanKeyboard.js` module similar to `src/modules/gobanGamepad.js`. What do you think?

hadim:
I've added support for keyboard navigation with `wasd` to move the cursor.

hadim:
@yishn are you interested in this PR? I am not sure I want to put more work on it if it ends up being rejected...

yishn:
Please do not put more work on it if you don't want to. Generally speaking, when opening pull requests, please make sure an issue exists beforehand to make sure your PR will be merged. Even if this PR is going to be rejected, I'm sure your work is not in vain. 

I'm still thinking about this; on one hand gamepad support is a very cool feature, on the other hand, this seems half-baked and I can't really contribute, since I don't have a gamepad to test it on. And maybe it's just too much for this project. 

I'm curious, is it possible to map gamepad controls to keyboard keys/mouse buttons? If that's possible, we only really need to provide keyboard vertex selection support.

hadim:
That feature is super useful/cool when playing on a TV. With the [following module](https://github.com/SabakiHQ/Sabaki/blob/f241daff59765d521e5261516ac4a91aae76d4c2/src/modules/gamepad.js), it's very easy to use and support various mappings/gamepad (it's also flexible in case we want to add unsupported gamepads.

Right now moving the cursor with a keyboard is also supported but the code structure might not be ideal. Before moving I wanted to be sure this is something interesting in your opinion.

Direct mapping of gamepads to the keyboard is not possible but can be easily done manually with the [gamepad.js module](https://github.com/SabakiHQ/Sabaki/blob/f241daff59765d521e5261516ac4a91aae76d4c2/src/modules/gamepad.js).

I see two options here:

- You can help me work on this PR by direct contribution and comments on the code.
- I'll give you the time to implement a complete keyboard support (with cursor navigation) and then I'll open a new PR that add gamepad support on top of it using the gamepad.js module.

Let me know your thoughts.

yishn:
The second option sounds good. I've opened an issue #456 regarding this, please subscribe there for updates. I can't say when I'm able to work on this, the analysis feature comes first for now.

hadim:
@yishn any news on this?

hadim:
Trying again to ask if you are still interested in that PR. I don't think it will add too much complexity to your code since it's just a single class doing a job and some callbacks raising the actions.

PR #476, opened 2018-11-15
Updated theme directory with additional user created theme.

JJscott:
Updated theme directory with additional user theme based on the BattsGo twitch stream. Although it's not meant to be a fully practical theme, it showcases what can be done with themes in general and is a good addition to diversify the current themes for v0.40.0


yishn:
I watched that stream yesterday 😆 It's amazing how much thought went in that theme, I didn't realize. Just one thing, can you append your entry to the bottom of the list instead?

JJscott:
No problem, the entry is at the bottom now. And thanks, it was a spirit of the moment thing. I kind of wanted to give it a go at some point, so it was a good opportunity.

PR #492, opened 2018-12-23
Add an option to change the Game Tree animation delay

### LINKED ISSUE #491: Add an option to change the Game Tree animation delay
## Motivation

The animation delay for a UI element that updates frequently may appear to be less responsive or even laggy. Being able to remove the delay may improve the overall feel of the component.

## Pros

The component feels more responsive with the delay set to 0ms

## Cons

With the delay being 0ms, the tree updates more frequently, and thus requires more computation power to appear smooth.

ohyou:
Resolve #491

## Issues

The `delay` variable in `src/components/GameGraph.js` was stored on file initialization and it was not accounting for the ability to be changed externally, despite the actual setting being present in the `settings`.

## Resolution

Explicitly update the `delay` variable on `componentDidUpdate`

### DIFF
--- a/src/components/GameGraph.js
+++ b/src/components/GameGraph.js
@@ -210,6 +210,9 @@ class GameGraph extends Component {
         if (showGameGraph !== this.props.showGameGraph) {
             setTimeout(() => this.updateCameraPosition(), 200)
         }
+
+        // In case it was changed externally
+        delay = setting.get('graph.delay')
     }
 
     getMatrixDict(tree) {
--- a/src/components/GameGraph.js
+++ b/src/components/GameGraph.js
@@ -6,7 +6,9 @@ const gametree = require('../modules/gametree')
 const helper = require('../modules/helper')
 const setting = remote.require('./setting')
 
-let delay = setting.get('graph.delay')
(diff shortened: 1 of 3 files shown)


yishn:
Wouldn't it better to just provide a checkbox saying 'Disable game tree animation'?

ohyou:
I've thought about that, but I thought that the dropdown would be more easily extensible.  
I can turn it into a checkbox, it shouldn't take long.

yishn:
True, but I don't think we need to be that extensible in this regard. Merry Christmas!

ohyou:
Introduced `graph.animation` toggle, but it only affects the delay right now. Should it completely disable all tree animations or it's ok like this?

yishn:
My bad, I was under the impression that there would be no animations when delay is set to `0`, but of course that's not true. Now I'm not sure what you want to achieve by setting delay to `0`? 

In the issue you mention that this makes the game tree feel more responsive, but it actually affects performance in a bad way and thus results in the game tree not being as responsive as it can be.

yishn:
I'm closing this due to inactivity.

PR #501, opened 2018-12-31
fix 499

dbosst:
Fixes area scoring for handicap games. Shows handicap column after komi, putting the # for white and - for black, (to make it obvious the row is added up).

Tested with handicap / non-handicap games.


yishn:
Thank you so much for adding the handicap column to the score drawer as well! It's really nice.

PR #503, opened 2019-01-01
Resizable Winrate Graph

dbosst:
Lets you resize the winrate graph like you can resize the move comment sidebar.

I have been zooming in/out too much to see more detail on the graph, so I made it resizable.

I ended up putting all the relevant code in the WinrateGraph.js rather than Sidebar.js, otherwise, it turned out the same component would end up having conflicting/confusing mouse event handling. Added a min. width in the settings.json, and it has a hard-coded 500 px max (should be plenty).

### DIFF
--- a/src/components/WinrateGraph.js
+++ b/src/components/WinrateGraph.js
@@ -3,20 +3,32 @@ const {h, Component} = require('preact')
 const helper = require('../modules/helper')
 const setting = remote.require('./setting')
 
+let winrateGraphMinHeight = setting.get('view.winrategraph_minheight')
 let winrateGraphHeight = setting.get('view.winrategraph_height')
 
 class WinrateGraph extends Component {
     constructor() {
         super()
 
+        this.state = {
+            sidebarWinrateGraphSplit: setting.get('view.winrategraph_height'),
+            sidebarWinrateGraphSplitTransition: true
+        }
+
         this.handleMouseDown = evt => {
             this.mouseDown = true
             document.dispatchEvent(new MouseEvent('mousemove', evt))
+        }
+
+        this.handleHorizontalResizerWinrateMouseDown = ({button}) => {
+            if (button !== 0) return
+            this.horizontalResizerWinrateMouseDown = true
         }
     }
 
     shouldComponentUpdate({width, currentIndex, data}) {
         return width !== this.props.width
+            || winrateGraphHeight !== this.state.sidebarWinrateGraphSplit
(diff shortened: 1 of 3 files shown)


src/components/WinrateGraph.js
```diff
@@ -25,6 +37,16 @@ class WinrateGraph extends Component {
         document.addEventListener('mousemove', evt => {
             if (!this.mouseDown) return
 
+            if (this.horizontalResizerWinrateMouseDown) {
+              evt.preventDefault()
```
yishn: Please adhere to 4-space indentation

src/components/WinrateGraph.js
```diff
@@ -3,20 +3,32 @@ const {h, Component} = require('preact')
...
+let winrateGraphMinHeight = setting.get('view.winrategraph_minheight')
 let winrateGraphHeight = setting.get('view.winrategraph_height')
 
 class WinrateGraph extends Component {
     constructor() {
         super()
 
+        this.state = {
+            sidebarWinrateGraphSplit: setting.get('view.winrategraph_height'),
+            sidebarWinrateGraphSplitTransition: true
```
yishn: Since we're not animating the winrate graph, I don't think keeping track of `sidebarWinrateGraphSplitTransition` is necessary?

These are very verbose state names. Can we simplify them a bit? E.g. just use `height` for `sidebarWinrateGraphSplit`.

src/components/WinrateGraph.js
```diff
@@ -3,20 +3,32 @@ const {h, Component} = require('preact')
...
+
+        this.handleHorizontalResizerWinrateMouseDown = ({button}) => {
+            if (button !== 0) return
+            this.horizontalResizerWinrateMouseDown = true
+        }
     }
 
     shouldComponentUpdate({width, currentIndex, data}) {
         return width !== this.props.width
+            || winrateGraphHeight !== this.state.sidebarWinrateGraphSplit
```
yishn: This is somewhat misleading. What you mean is probably:

~~~js
shouldComponentUpdate({width, currentIndex, data}, {sidebarWinrateGraphSplit}) {
    ...
        || sidebarWinrateGraphSplit !== this.state.sidebarWinrateGraphSplit
    ...
~~~

This will check whether the `sidebarWinrateGraphSplit` has changed or not.

yishn:
Thank you so much for your pull request! This is probably the best new year gift ever. Happy New Year!

PR #504, opened 2019-01-02
Resize gameboard when toggling show sidebars

dbosst:
When toggling right or left sidebars, the board doesn't get resized & redrawn to fit the window.
This fixes that.


yishn:
Thanks!

PR #505, opened 2019-01-02
Fix for menu checkbox state

dbosst:
When you clicked on either "View-> Show Game Tree" or "View-> Show Comments" the checkboxes weren't updating immediately.

This is fixed in this patch, which adds the missing state tracking code.

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -331,6 +331,8 @@ class App extends Component {
             'view.show_move_colorization': 'showMoveColorization',
             'view.show_next_moves': 'showNextMoves',
             'view.show_siblings': 'showSiblings',
+            'view.show_graph': 'showGameGraph',
+            'view.show_comments': 'showCommentBox',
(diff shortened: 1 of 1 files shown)


yishn:
Which OS do you use? On my side (Windows 10) the menu item checkboxes will update immediately.

The tracking of state where you added your code are for real time updates. Changes in `setting` will be tracked application-wise. If we add your code there the following unwanted behavior occurs:

1. Open Sabaki
2. Press <kbd>Ctrl/Cmd</kbd>+<kbd>Shift</kbd>+<kbd>N</kbd> to open a new window
3. Click 'View' > 'Show Game Tree'
4. Game tree shows/hides in both windows simultaneously

We want each window to track their own game tree/comments visibility states.

dbosst:
Ubuntu 16. Ah I see I didn't know it worked like that.
At the moment when you click show game tree then click view again it stays stuck unless you click on something like show next move (something above it in same submenu)

PR #506, opened 2019-01-03
Animated analysis variations - continuous updating

dbosst:
For only animate the variation mode: extends the animation of the current analyzed variation when more (deeper) moves arrive for the same variation (like a ladder) -- a different variation for the same root move won't be updated, only for the same variation already displayed.

Cons: Might be of limited use since its not very likely the current variation is always continued for every single move... so I created it separately from the other pull request for your opinions on whether its worthwhile.

### DIFF
--- a/src/components/Goban.js
+++ b/src/components/Goban.js
@@ -150,20 +150,17 @@ class Goban extends Component {
                         let vertex = variation[0];
                         if (vertex == null) return
                         let analysisline = this.props.analysis.find(x => helper.vertexEquals(x.vertex, vertex)) || {}
-                        if (analysisline == null)
-                            return
+                        if (analysisline == null) return
                         let nextvariation = analysisline["variation"]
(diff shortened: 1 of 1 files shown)


yishn:
This sounds like a nice feature, but why is it only for no instant replay mode only?

dbosst:
If its implemented for the animated version, you might never see it play out the entire animated line... the engine would just start playing the first N moves then jump to the next variation from the beginning ... I can't think of a way to make it work well. It feels like this way there is two different alternatives (instant for quick review, and animated for imitating you slowly placing the stones yourself)

yishn:
Wait, but this is for the animated version, right? Am I understanding it wrong?

dbosst:
=( oops I mixed up the wording, and confused myself. It's only for the animated version. The other commit would have effectively implemented the same thing.

yishn:
I'm closing this due to inactivity.

PR #507, opened 2019-01-03
Continuous updates for analysis variations in instant replay

dbosst:
For instant replay mode only: continuously display the last variation for the move, changing it at most every replay_interval.  It is a setting right now as default true: "board.variation_instant_replay_update_continuously"

Also, added keyboard shortcut of Ctrl+Space (or Meta+Space) that has the following functionality:

When the setting board.variation_instant_replay_update_continuously == true .... When viewing the variation you can pause/unpause continuous updates for the variation using the keyboard shortcut. When you move the mouse off the variation (stopping it) and go back to the same or a different variation the continuous updates are reset and will display.

When the setting board.variation_instant_replay_update_continuously == false, the keyboard shortcut does nothing, and reverts back to the behaviour before the patch.

This gives you time time to study variations: if you find one interesting enough to ponder it for a while you can hit ctrl-space to pause before the next variation comes, and when you are done pondering you can hit ctrl-space to see if there is a new variation for the move, without needing to move the mouse off and back on again.

### DIFF
--- a/src/components/Goban.js
+++ b/src/components/Goban.js
@@ -16,8 +16,10 @@ class Goban extends Component {
         this.handleVertexMouseMove = this.handleVertexMouseMove.bind(this)
         this.handleVertexMouseEnter = this.handleVertexMouseEnter.bind(this)
         this.handleVertexMouseLeave = this.handleVertexMouseLeave.bind(this)
+        this.handlePauseContinuousUpdates = this.handlePauseContinuousUpdates.bind(this)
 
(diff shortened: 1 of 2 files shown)


yishn:
As you said, you can do this by moving the mouse off and back on again, so I don't think this feature is necessary. I don't want to trade in the extra complexity for this. Besides, the keyboard shortcut has issues on macOS, where <kbd>Cmd</kbd>+<kbd>Space</kbd> is normally used to open Sportlight search.

PR #508, opened 2019-01-05
Gtp console logging, fix #272

### LINKED ISSUE #272: Seki
https://imgur.com/a/OkMhj
Incorrect counting 2 white points. Maybe can make empty area on click?

dbosst:
Console logging with winston using events.  Logging is gathered using events to reconstruct the log, rather than modifying the GTP module to extract the raw GTP data (like this we can track the metadata better, i.e., multiple matches, engine names, player color)  All the important data should be included, stdio/stdin/stderr are included (including analysis in the info move lines).

I've only tested OK on my own system, Ubuntu 16.04. It really needs more testing and review.

Some notes on implementation details:
1. All logging is done to one, combined log file (same one used by all new windows)
2. The combined log file is still parseable even with simultaneous matches (see below example)
3. Three new settings in the settings.js. Logging enabled & log file path are in the GUI under Engines. The other setting for timestamps, 'gtp.console_log_timestamps' is only available through editing the config.json manually.
4. A few extra custom lines in the log were added for letting events of the engine status ("Sabaki: Engine Started / Suspending / Stopped / Connection Failed")
5. If a bad logging file path is given on startup, a warning dialog shows up.
6. The modified code for the GUI preferences isn't pretty (and I didn't want to write a whole other class for one item). If it needs to be redone let me know.

An example snippet from simultaneous two different matches in the logfile: (after the time stamp "2.W" refers to 2nd window White player, "1.B" refers to 1st window Black player):

```
[2019-01-05T18:28:24.790Z] 1.B <LZ-15B-T0-10> Playouts: 67, Win: 46.72%, PV: E3 F4 Q4 C4 G2 O17 F17 C13 K16 R17 Q17 R16 Q15 Q18 P18 R18 O18 R14 N17
[2019-01-05T18:28:24.936Z] 2.W <PZ-V160> info move Q16 visits 1 winrate 5491 prior 8360 order 0 pv Q16  info move Q4 visits 1 winrate 5483 prior 453 order 1 pv Q4  info move R16 visits 1 winrate 5466 prior 16 order 2 pv R16  info move D16 visits 1 winrate 5462 prior 474 order 3 pv D16 

```


dbosst:
At the moment I can only run this when doing "npm start" -- I can't figure out how to include it in the bundle for building.

yishn:
Does `npm run build` not work?

dbosst:
> Does `npm run build` not work?

npm run build [b]Works[/b] for me (with 3a0b505).

dbosst:
Yes, `npm run build` works. (as of 8230348)

yishn:
Is there a reason why `logger.js` has to be a remote module? If it's a renderer module, it will be bundled by webpack automatically like all the other modules in the `modules` folder.

dbosst:
> Is there a reason why `logger.js` has to be a main module required by `electron.remote`? If it's a renderer module, it will be bundled by webpack automatically like all the other modules in the `modules` folder.

The last commit 692794a moves logger.js out of modules into the main folder where it belongs, so it is included automatically in webpack and no longer requires that change to the package.json.

logger.js should be treated like setting.js which is just one instance & global.  It can't be a renderer module since we only need one global instance of it. Another reason for it to be a main module is that we need to validate that the settings for the log file path is valid at the app load, and warn the user if the log file path specified is invalid (this is important to know).

yishn:
When you say

> logger.js should be treated like setting.js which is just one instance & global.

will there be a problem when you have multiple Sabaki processes running? If not, I don't see any issues with it being a renderer module. You can access the `setting` module by `remote`. Checking the validity of the log file path can be done at engine attachment, not at app load.

dbosst:
85c3a43 will validate the log file on every engine attachment now instead of app start.
The only difference in behavior: you get a message on every engine attachment until you resolve the bad path (it also happens on exit & engine detach ... since for some reason attachEngines is called in those cases ... I don't think this can be avoided on my side).

yishn:
> multiple windows would each incorrectly try to create a new logger, each of which would try open additional file handles to the SAME log file when attaching.

That would also be the case when you have multiple processes of Sabaki running, no? As a main module, each process would require `logger` once, but once you have multiple processes, wouldn't they all have a file handle to the same log file open?

dbosst:
> > multiple windows would each incorrectly try to create a new logger, each of which would try open additional file handles to the SAME log file when attaching.
> 
> That would also be the case when you have multiple processes of Sabaki running, no? As a main module, each process would require `logger` only once, but once you have multiple processes, wouldn't they all have a file handle to the same log file open?

Hmm... good point, I havent considered the multiprocess case where there are multiple different main processes, since I usually just click file->new window.  There are two different concerns thinking about this:

1. When I tested it in Linux there were no file locking issues with winston even with multiple handles open, which means there _probably_ wouldn't issues be implementing it as a renderer (the distinct downside being there being multiple file handles and more memory usage for multiple windows with a main process).  As for Windows/Mac I don't know what its behavior would be.. I need to get it on a Windows/Mac machine to test this behavior. If it tests OK, I'll implement it as a renderer as you requested.

2. Aside from file locking, another concern is that logger currently logs the Window number relevant to the single main process so that people can run multiple engines and differentiate them.  If we have multiple different main processes of Sabaki, then in the combined log file, the window number won't be unique any more (since its relative to each process).  In order to make the logging unique with the multiprocess then we need to do something like also add main process id with the window number:
The log file then would look like: (ie. process id = 25440):
`[2019-01-05T18:28:24.790Z] 25440 1.B <LZ-15B-T0-10> Playouts: 67, Win: 46.72%, PV: E3 F4 Q4 C4 G2 O17 F17 C13 K16 R17 Q17 R16 Q15 Q18 P18 R18 O18 R14 N17`
If you have any other ideas for 2. other than adding the `pid` let me know.

dbosst:
Ok, the last couple of commits raised some inconsistent behaviour from synchronization between windows of the logging settings and the logger state (multiple windows, multiple processes), discussed below in the changes:

1. Moved `logger` from main into as a renderer module, loaded by `App`.  Each renderer window (regardless of multi-process or not) will have its own logger/file handle to the log file.  There is some inconsistent behavior for log file path and logging enabled options:
1.1. For changes to **logging enabled/disabled**: logging starts/stops for all renderer windows common to the `main` process where the change occurred. In renderer windows belonging to other `main` processes (multiprocess case) this **change is not updated all** since settings.json is only read at startup.  Handling  this (multiprocess case) would require IPC using something like sockets with a separate broker (more complexity than its worth IMO).
1.2. For changes to the **log file path**: For a specific window, if you have engines attached, and if you then change the preferences for the log path in that window, this will cause an immediate switch to the new log file (without needing to detach/attach) -- I believe this is the desired behaviour, but **this can be changed** easily if you think otherwise. This change is **only immediate for the current window**.  If there are other windows under the same main process, their logger's log path won't be updated **until** they attach/detach/check/change settings (**this case can be changed** with more IPC code, i.e. IPC from renderer to main to renderers)). But, for other windows belonging to a different main process, their logger's log path **won't be updated at all** (this issue has the same problems implementing as the multiprocess case of `1.1`

2. Re-added exception handling to `main`. In case we need to quit unexpectedly, this should try to flush/close the loggers for every renderer window belonging to the `main` process.

3. Change to the log file format: dropped window numbers, and now use the renderer's OS process id as the identifier for the match in the log.  This covers all cases of uniqueness multiple windows or multiprocess.  Example (11621 and 13090) are process ids for two renderer windows:
```
[2019-01-09T04:09:51.457Z] 11621 : W <LZ-15B-T0-10> info move R3 visits 480 winrate 37 prior 1763 order 0 pv R3 Q3 R4 R6 R5 Q5 S6 R2 S2 S5 Q2 S3 R1 S4 R2 S7 Q6 R7 O3 O6 M4 R14 C17 D17 C16 C14 C15 D15 B14 info move R17 visits 393 winrate 36 prior 1456 order 1 pv R17 Q17 R16 R14 R15 Q15 S14 R18 S18 S15 Q18 S17 R19 S16 R18 S13 Q14 R13 O17 O14 M16 R6 C3 D3 C4 C6 C5 D5 B6]
[2019-01-09T04:09:51.552Z] 13090 : W <LM-15B-GXAA> Playouts: 885, Win:  0.75%, PV: C14 F17 B16 C13 D13 C12 D14 C17 D12 C10 C11 B11]```

**Comment by yishn** [comment]
So, after a cursory search, almost all logging libraries have issues with multiple processes logging into the same file; simultaneous logging by multiple processes will result in data loss in the best case, or corruption in the worst case.

Another idea would be to let the user specify a log directory rather than a file. Then we can create a new log file (with timestamp, engine names, counter in its file name) for every "engine session". Of course we need to define what an "engine session" is.

**Comment by dbosst** [comment]
> So, after a cursory search, almost all logging libraries have issues with multiple processes logging into the same file; simultaneous logging by multiple processes will result in data loss in the best case, or corruption in the worst case.
> 
> Another idea would be to let the user specify a log directory rather than a file. Then we can create a new log file (with timestamp, engine names, counter in its file name) for every "engine session". Of course we need to define what an "engine session" is.

We can open a new log file on either (1) on new game trees, e.g. with every File->new or File->Open, or (2) on new renderer windows.  For either, log file naming would be "YYYY-MM-DD-hh-mm-ss_PID.log".  In either case we are going to end up with a lot of log files (do we need to copy what syslog's behaviour and archive old ones after a week?).  If we have a lot of log files, we might as well try using (1), I think that would be the expected behaviour from users in this case.

**Comment by dbosst** [comment]
Latest commit:
1. Implements (1) above, to rotate to new log file for each file->new, file->open, and, obviously also for new renderer windows and multiprocess' renderer windows). Implemented the changes to reflect now choosing a directory and not a file in the GUI and in logger code (this resulted in simpler code in some areas).
2. Fix for timestamps so shows local time for winston in log file
3. Log file naming and log format example (12435 is the process id):
`sabaki_2019-01-09-13-01-43_12435.log`
`[2019-01-09 13:01:51.737] W <LZ-40B-T10-30> info move C17 visits 1 winrate 136 prior 1933 order 0 pv C17`

Behaviour notes:
1. On window close, any last few lines may not be written to the logger (we don't want to hold up the window when the user wants to quit, do we?)
2. The synchronicity behaviour/issues with preferences across renderer windows or multiprocess that I mentioned in the previous post still apply: the logging folder won't be updated in those cases. These are edge cases that frankly I don't think we should be concerned with.

File: src/components/App.js [hunk 1]
```diff
@@ -452,6 +468,8 @@ class App extends Component {
 
         await this.loadGameTrees([emptyTree], {suppressAskForSave})
 
+        try {logger.rotateLog()} catch (err) {}
```
yishn: Why not call `rotateLog()` in `loadGameTrees()`? That way you don't need to call it in `newFile()`, `loadFile()`, and `loadContents()` separately.

src/modules/logger.js
```diff
@@ -0,0 +1,205 @@
...
+
+exports.unload = function() {
+    try {
+        logger.close()
+    } catch (err) {}
+}
+
+exports.close = function() {
+    try {
+        logger.close()
```
yishn: What does `logger.close()` do exactly? I can't find this method in [winston's documentation](https://github.com/winstonjs/winston). Can you point me to some reference?

src/components/App.js
```diff
@@ -2038,6 +2060,8 @@ class App extends Component {
             return
         }
 
+        try {logger.loadOnce()} catch (err) {}
```
yishn: This seems like an arbitrary place to load something once. Can we avoid calling `logger.loadOnce()` altogether? Maybe we can call `loadOnce` automatically (in `logger` itself) when we call `writeToLogFileGTP()`?

src/components/App.js
```diff
@@ -246,6 +247,20 @@ class App extends Component {
             }
         })
 
+        ipcRenderer.on('updateLogger', (evt, ...args) => {
```
yishn: The only place that sends this ipc event is in `PreferencesDrawer`. We can just execute this code there, can't we? That way, we don't need to make an ipc call.

dbosst: We can't. The logger instance is local to App.js, and since logger.js a module it isn't possible to access the logger instance in PreferencesDrawer, since that would need a remote.require(), which is only possible if logger.js wasn't a module.

src/components/App.js
dbosst: The placement has entirely to do with the desired behaviour of to give the user a warning/error of an invalid log folder path when an engine is attached.  I think it would be awkward if not bad if that warning fired when we actually tried writing to the log.  If you want to get rid of this warning message all together, however, we could move the loadOnce() into writeToLogFileGTP, by doing the load inside an `if (logFileNameGTP==null)`

yishn: I see, then I prefer it to have it here.

src/components/App.js
yishn: Why not? You can use `require` like all the other modules.

src/components/App.js
dbosst: Hmm, good point. This is fixed in ee14a8a

src/modules/logger.js
dbosst: `close()` calls `clear()`, which closes all the logger transport streams -- all in all the purpose of it in our code is to **make sure the data is flushed before we quit**.

There isn't  thorough documentation for all of winston, so I had to go through winston's code to answer you more definitively:

`close()` calls `clear()`, which closes all the streams with Node's `stream.unpipe()` -- from my understanding, removing this writable stream with unpipe makes sure any data in the stream's buffer is flushed to disk before we actually quit the application.

**The only reason for using `close()` rather than `clear()` is it's easier to understand than `clear()` from first glance** (I've seen it used interchangeably with `clear()` in winston examples).  `clear()` _sounds_ like it is clearing the buffer, but that is not what it does (terrible naming IMO) -- it just removes all the transport streams for the logger instance using `unpipe()`.

`.close()` real intended use seems to be for the case when you have a container of different winston logger instances, then you can use events to manage them: `close()` fires a 'close' event an event to let the manager know we did a `clear()` on that logger instance

**On a related note**: at the moment we don't actually bother to remove the logger until we need to actually start a new logging file with `updateLoggerPath()`. There is a very good reason for this: the logger is async, so its always possible to be writing data past a Sabaki call for a new board with `loadGameTrees()`, especially if the user didn't manually detach the engines first, so we need to keep the current logger open as long as possible in that eventuality. (In testing I saw this happen, which is why it is like it is rather than using a much more complicated callback system).

Here are the code line references in the source to `close` & `clear()`:

`close()` calls `clear()`
https://github.com/winstonjs/winston/blob/64344d12eace394d9e7c8bf926f9f2663efc8640/lib/winston/logger.js#L348

`clear()` calls `unpipe()`
https://github.com/winstonjs/winston/blob/64344d12eace394d9e7c8bf926f9f2663efc8640/lib/winston/logger.js#L338

src/components/App.js
dbosst: Oops I put it in the wrong place, 86147a5 should definitely fix it =)

src/components/App.js
dbosst: Ok, then there is nothing else I can do then.

src/components/App.js
dbosst: Doing a normal require doesn't give me access to the logger.js instance that is created in App, which holds state information and the actual winston.logger instance.  We can only get away with doing a normal require as it is in PreferencesDrawer since we don't actually need to use any of that above: only the logger.js helper functions for validating the logging path are needed.

yishn: > Let me put it another way, doing a normal require in PreferencesDrawer and trying to update without IPC would be the same as futilely trying to access the logger from a different process (they are different instances)

It's a common misconception that Node.js will load a module on each `require`. Actually Node.js [will cache the module object](https://nodejs.org/api/modules.html#modules_caching) and reuses it on each subsequent `require`.

If it puts your mind at ease, you can also access the "right" logger.js instance by `sabaki.modules.logger`.

dbosst: I see! Thanks for telling me that.

OK. I've changed it in the code.  This is resolved now with 08fbd0a

src/components/App.js
```diff
@@ -256,6 +261,7 @@ class App extends Component {
             setTimeout(() => {
                 if (this.askForSave()) {
                     this.detachEngines()
+                    try {logger.close()} catch (err) {}
```
yishn: Please try to eliminate all `try` blocks with `logger` calls by moving them into the `logger` module (if necessary at all).

src/modules/logger.js
```diff
@@ -0,0 +1,205 @@
...
+        exports.updateLogFilePathGTP()
+    }
+}
+
+exports.rotateLog = function() {
+    // On next engine attach, we will use a new log file
+    logFileNameGTP = null;
+}
+
+exports.unload = function() {
```
yishn: This function is not being used.

src/modules/logger.js
```diff
@@ -0,0 +1,205 @@
...
+
+        GTPText = ((enginestream.sign === 0) ? "B" : "W" ) +
+            (enginestream.name != null ? " <" + enginestream.name + ">" : " <>") + " " +
+            GTPText
+
+        logger.log('info',GTPText)
+    } catch (err) {}
+}
+
+exports.clearLoggers = function() {
```
yishn: This function is not being used.

src/modules/logger.js
```diff
@@ -0,0 +1,205 @@
...
+const setting = remote.require('./setting')
+const dialog = require('./dialog')
+const winston = require('winston');
+const path = require('path')
+
+let logger = null
+let logFileNameGTP = null
+let staleLogFilePath = false
+
+exports.writeToLogFileGTP = function(enginestream) {
```
yishn: First of all, I would like to clean up this module by renaming lengthy function names like this one to something shorter. Probably `write` will even suffice here, since this module is all about logging GTP data from the console. In light of this, it's probably best to rename `logger` into `gtplogger`.

dbosst: Let me know if the new names are OK. The new commit 6b159c6 should resolve this and all all of the above requested changes.

src/modules/gtplogger.js
```diff
@@ -0,0 +1,195 @@
+const {remote} = require('electron')
+const fs = require('fs');
+const setting = remote.require('./setting')
+const dialog = require('./dialog')
+const winston = require('winston');
+const path = require('path')
+
+let gtplogger = null
```
yishn: Maybe call this variable something else to avoid confusion. `winstonLogger`, or simply `logger`? On another note, can't we call `winston.createLogger` right here?

src/modules/gtplogger.js
```diff
@@ -0,0 +1,195 @@
...
+        }
+        return false
+    } else {
+        return true
+    }
+}
+
+exports.loadOnce = function() {
+    if (gtplogger == null) {
+        if(setting.get('gtp.console_log_timestamps')) {
```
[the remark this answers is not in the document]
dbosst: I suppose not. If they don't want it, they can always use a simple tool like "cut" to trim the first 27 characters. I have removed it.

src/components/App.js
```diff
@@ -2051,6 +2057,12 @@ class App extends Component {
                 this.attachedEngineSyncers[i] = syncer
 
                 syncer.controller.on('command-sent', evt => {
+                    gtplogger.write({
+                        'stdin': evt.command.name,
+                        'args': evt.command.args,
+                        'sign': this.attachedEngineSyncers.indexOf(syncer),
+                        'name': engines[i].name
+                    })
```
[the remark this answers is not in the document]
dbosst: Done

src/components/drawers/PreferencesDrawer.js
```diff
@@ -184,13 +185,23 @@ class PathInputItem extends Component {
         }
 
         this.handleBrowseButtonClick = evt => {
-            dialog.showOpenDialog({
-                properties: ['openFile'],
-            }, ({result}) => {
-                if (!result || result.length === 0) return
+            if (this.props.chooseGTPLogDir != null) {
```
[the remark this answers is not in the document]
dbosst: Very good. Done.

src/modules/gtplogger.js
dbosst: Yep. Done.

dbosst:
My last few commits should have resolved your previous review.
Since simplifying the object passed to write(), I changed the logging line a little bit so that it includes the channel 'type' like: 'err', 'in', 'out', 'meta'. Here are some examples:

> [2019-01-15 16:36:38.206] B <LZ-P1> (meta) : Engine Started
[2019-01-15 16:37:17.921] B <LZ-P1>  (out) : play D17
[2019-01-15 16:37:16.415] B <LZ-P1>   (in) : lz-genmove_analyze B 100
[2019-01-15 16:37:11.610] B <LZ-P1>  (err) : Thinking at most 37.5 seconds...

In reality the message text, "Engine", "play", "lz-genmove", "Thinking", is all aligned (github removing multiple leading spaces)

src/modules/gtplogger.js
```diff
@@ -0,0 +1,195 @@
...
+}
+
+timestamp = function() {
+    let now = new Date();
+    let t = {};
+    t['mon'] = 1 + now.getMonth();
+    t['day'] = now.getDate();
+    t['hour'] = now.getHours();
+    t['min'] = now.getMinutes();
+    t['sec'] = now.getSeconds();
```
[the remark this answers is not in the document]
dbosst: Done.

src/modules/gtplogger.js
```diff
@@ -0,0 +1,195 @@
...
+const dialog = require('./dialog')
+const winston = require('winston');
+const path = require('path')
+
+let gtplogger = null
+let filename = null
+let stalePath = false
+
+exports.write = function(stream) {
+    let GTPText = ""
```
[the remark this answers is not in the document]
dbosst: ok

src/components/drawers/PreferencesDrawer.js
```diff
@@ -578,6 +608,12 @@ class PreferencesDrawer extends Component {
 
             setting.set('engines.list', engines)
 
+            if (sabaki.modules.gtplogger.validate()) {
+                if (sabaki.attachedEngineSyncers.some(x => x != null)) {
+                    sabaki.modules.gtplogger.updatePath()
+                }
+            }
+
```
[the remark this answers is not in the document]
dbosst: For the first point, I've made it so it re-opens the engines tab to force the user to fix it.

Regarding the second point, we need to check if engines are attached. In `updatePath` the log file gets created on the disk, so if we call it from Preferences with no engines attached we will end up creating empty log files.
I've rearranged the code so this gets checked in updatePath now, by letting updatePath take a true/false state, so we know whether we are calling it from attaching Engines in `App` or from `PreferencesDrawer`.  This is necessary for the correct behaviour to make sure we always `validate` in either case.

src/modules/gtplogger.js
```diff
@@ -0,0 +1,156 @@
...
+        }
+        return false
+    } else {
+        return true
+    }
+}
+
+exports.loadOnce = function() {
+    if (exports.validate()) {
+        exports.updatePath()
```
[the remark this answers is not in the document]
dbosst: Done. When refactoring I also spotted several places to simplify and clean up `updatePath` and `validate`.

src/components/drawers/PreferencesDrawer.js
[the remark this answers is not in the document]
yishn: No, I don't want to give `updatePath` a flag whether it's being called from `App` or `PreferencesDrawer`. It doesn't make any sense design-wise. What I meant by the second part is: Instead of

~~~js
if (sabaki.modules.gtplogger.validate()) {
    if (sabaki.attachedEngineSyncers.some(x => x != null)) {
        sabaki.modules.gtplogger.updatePath()
    }
}
~~~

we can do

~~~js
if (sabaki.attachedEngineSyncers.some(x => x != null)) {
    if (sabaki.modules.gtplogger.validate()) {
        sabaki.modules.gtplogger.updatePath()
    }
}
~~~

thus failing silently if there are no engines attached. This way, `updatePath` can always call `validate` and there's no need for a flag.

dbosst:
8a85a42 .. 2273f05 looks good.

yishn:
Thanks for checking my refactoring 👍 I still plan to restructure the CSS before merging.

PR #530, opened 2019-03-16
Update engines.md

bdrose:
Updated Pachi link and description


bdrose:
Hopefully this is correct now.

PR #540, opened 2019-03-26
Initial swap colors function

### LINKED ISSUE #538: Interchange colors / Switch colors
An **Interchange colors** or **Switch colors** command in the **Tools** menu would be very useful for the purpose of comparing positions. The command would be used to interchange black and white stones, that is switch black stones to white and vice versa.  (Similar positions can arise but with the colors reversed. It would be good to be able to switch the colors in any position to make it easier to compare positions, and also to help identify such positions.)

fohristiwhirl:
Some comments:

* Perhaps this file shouldn't be called rotation any more, or some functions should be moved out of it.
* Currently it just updates B, W, AB, and AW. Should we do metadata like PB, BR etc?

Closes #538

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -1733,6 +1733,14 @@ class App extends Component {
         this.setCurrentTreePosition(newTree, treePosition, {clearCache: true})
     }
 
+    swapColors() {
+        let {treePosition, gameTrees, gameIndex} = this.state
+        let tree = gameTrees[gameIndex]
+        let newTree = rotation.swapTreeColors(tree)
+
+        this.setCurrentTreePosition(newTree, treePosition, {clearCache: true})
+    }
+
     copyVariation(tree, treePosition) {
         let node = tree.get(treePosition)
         let copy = {
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -1733,10 +1733,10 @@ class App extends Component {
         this.setCurrentTreePosition(newTree, treePosition, {clearCache: true})
     }
 
-    swapColors() {
+    invertColors() {
         let {treePosition, gameTrees, gameIndex} = this.state
         let tree = gameTrees[gameIndex]
-        let newTree = rotation.swapTreeColors(tree)
+        let newTree = rotation.invertTreeColors(tree)
 
         this.setCurrentTreePosition(newTree, treePosition, {clearCache: true})
     }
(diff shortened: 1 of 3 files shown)


yishn:
> Perhaps this file shouldn't be called rotation any more, or some functions should be moved out of it.

Yeah, `boardTransformation` suits more.

> Currently it just updates B, W, AB, and AW. Should we do metadata like PB, BR etc?

That's a good question. Probably not... After all, we're also not updating all coordinates referenced in the comments after a rotation/flip. I'm still doubtful whether anyone actually uses the board transformation functions to actually transform the SGF, or if they just want to look at the board from another angle temporarily.

fohristiwhirl:
I think rotation has some merit for game recorders. The others are probably less useful.

src/components/App.js
```diff
@@ -1733,6 +1733,14 @@ class App extends Component {
         this.setCurrentTreePosition(newTree, treePosition, {clearCache: true})
     }
 
+    swapColors() {
```
yishn: Regarding nomenclature, I think we should stick with "invert colors" to have it the same as digital painting/photo editing software.

src/modules/rotation.js
```diff
@@ -5,6 +5,13 @@ const arrowish = ['AR', 'LN']
 const alpha = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'
 const alphaRev = [...alpha].reverse()
 
+const colorSwapMap = {
+	'AB': 'AW',
```
yishn: Be careful of the indentation. You've slipped in some tabs inside this file.

fohristiwhirl:
Um, there's the issue that changing the colors in the middle of a game should change who is to play next, I don't immediately know how to do that.

yishn:
> Um, there's the issue that changing the colors in the middle of a game should change who is to play next, I don't immediately know how to do that.

I will look into that, thanks for the heads-up.

fohristiwhirl:
I guess most of the time Sabaki infers who is to play next from the previous move, so there's no issue?

yishn:
@fohristiwhirl Uhm... Are you saying there is an issue or not? Theoretically, there shouldn't be, yes.

fohristiwhirl:
Yeah I think the only issue is if you set a handicap and then immediately flip the colours, then play a stone, that stone will be white, but this is fairly unimportant.

fohristiwhirl:
I actually found a valid reason for flipping colours today... I had an old game from the classic era where white played first, and wanted to feed it to Leela to analyse, but since Leela is trained with komi, white playing first would make it even more wrong than normal...

So flipping the colours made sense.

PR #542, opened 2019-04-03
Fix blurred background

ParmuzinAlexander:


yishn:
I think I'd rather have a slightly blurry background image than a background image that's not centered...

PR #548, opened 2019-04-13
Pause loop for sending analysis commands during generating moves.

dbosst:
This will preserve the 'analysis state' better so when Sabaki finishes generating one or more moves it will start analysis again.

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -1325,7 +1325,9 @@ class App extends Component {
             n = tree.get(n.parentId)
         }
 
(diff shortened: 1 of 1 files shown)


yishn:
In my fix, Sabaki will also start analysis again after Sabaki finishes generating one or more moves. The 'analysis state' is being preserved. Are there any other advantages to this PR? Otherwise, I'd prefer my fix.

dbosst:
Ah I didn't think it would do so when I simulated it in my head - I forgot the call to startAnalysis in setTreePosition.

Your's is perfect. I'm closing.

PR #551, opened 2019-04-14
Winrate graph invert option

fohristiwhirl:
I don't know if you'll like this, but this adds an option to invert the winrate graph so that the data line itself tracks Black's winrate, which I think is more intuitive, Black being player 1 and all.

I didn't fully understand the render function so hacked at it as best I could. If you decide to accept it, I'd recommend checking my changes make sense...

Also, there's the minor issue that the change in the preferences won't take effect until the graph is redrawn. In fact, one could argue it doesn't need a preferences item at all, with the `settings.json` item being sufficient.

### DIFF
--- a/src/components/WinrateGraph.js
+++ b/src/components/WinrateGraph.js
@@ -66,6 +66,13 @@ class WinrateGraph extends Component {
     }
 
     render({width, currentIndex, data}) {
+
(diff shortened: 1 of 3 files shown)


yishn:
Looks useful, thanks!

PR #561, opened 2019-04-27
Fix #560

### LINKED ISSUE #560: Flatten forgets the current player
After the Flatten operation, the current player is always Black, even if it was White's turn in the current position.

dfannius:
I'm not that familiar with the Sabaki codebase so I did my best but I don't know if this fix is idiomatic.

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -1835,6 +1835,7 @@ class App extends Component {
         if (gameIndex < 0) return
 
         let board = gametree.getBoard(tree, treePosition)
+        let player_sign = this.getPlayer(tree, treePosition)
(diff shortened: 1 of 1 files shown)


src/components/App.js
```diff
@@ -1835,6 +1835,7 @@ class App extends Component {
         if (gameIndex < 0) return
 
         let board = gametree.getBoard(tree, treePosition)
+        let player_sign = this.getPlayer(tree, treePosition)
```
yishn: Please use camelCase.

dfannius: Righto!

PR #572, opened 2019-05-10
Faster board transformations

yishn:
~~Instead of transforming the entire game tree, we will now store what transformations have been made in a root node SGF property, and we transform the board in place.~~ We will only transform the board in place, not save the transformation inside the SGF at all.

This way, not only the transformations will be faster, but we can also transform the coordinates as well, leaving coordinate references in the comments intact. Furthermore, we will not have to alter the game history, which will be helpful in collaborative scenarios in the future.


fohristiwhirl:
Course, this makes the transformation invisible to other SGF editors, SGF database tools, etc.

yishn:
> Course, this makes the transformation invisible to other SGF editors, SGF database tools, etc.

That's right. After some thought, I've also decided not to persist the transformation inside the SGF at all. So it won't have any effect on the SGF; it's an application setting.

fohristiwhirl:
Currently "Copy ASCII diagram" does not take transformations into account.

PR #665, opened 2020-03-18
Add option to select the coordinate type

ebifrier:
This allows you to choose from three different formats for displaying coordinates.

One is the same display format as before.
The other one is commonly used in Japan. The upper left is (1,1), and the numbers increase in the right and lower directions.
The last one is the format used in Chinese foxweiqi. I don't know if this is standard, but added it just in case.

IMG_1377


yishn:
Thanks for your pull request! I like the feature to toggle between different coordinate systems. I would prefer if this isn't a setting you can change in the Preferences drawer, but instead a submenu inside the 'View' main menu.

Do you have any evidence for your so-called Japanese style coordinates? I find it hard to believe that both axes are labeled by a number. Also, what I find common is that some people use 一二三四... instead of ABCD...

ebifrier:
As you say, the exact coordinate display format is 一二三四...

However, adopting this format requires two multi-byte characters to be displayed vertically, which makes the program complicated.
Also, there is a possibility of garbled characters outside of the Japanese-speaking world, so I didn't implement it.

ebifrier:
If you think this feature is necessary, I'll try to implement it.

ebifrier:
For now, I moved the settings to the submenu in the 'View' main menu.

Note: Shortcuts that corresponded to "Show Coordinates" were deleted due to the changed menu.
I didn't know the right way, so I'd like you to add it as needed.

Thank you for your advice !

yishn:
Thanks, I think it's best to assign 'Don't Show' the keyboard shortcut, and change the behavior, so it toggles the visibility instead of hiding it.

GopherAl:
@ebifrier It would be fantastic if you could also include the coordinate system I requested help with in issue 4 in SabakiHQ/Shudan. [https://github.com/SabakiHQ/Shudan/issues/4#issuecomment-475250765]. @yishn suggested some code.
Perhaps a better name for this coord system would be **Descriptive coordinates** or **Natural coordinates**. See below for details:

When using Sabaki I would like to be able to choose the alternative coordinate system "Simple coordinates". See Senseis Library reference at https://senseis.xmp.net/?Coordinates#toc7

Using these coordinates, on the 19x19 goban the x and y axes are both labelled 1 2 3 4 5 6 7 8 9 X 9' 8' 7' 6' 5' 4' 3' 2' 1'   ("X" is of course the Roman numeral for 10, widely used in English-speaking countries.)

Examples (default coords first, simple coords second): D4=44; C16=34'; K10=XX; R17=3'3'.
It would only be appropriate to use these coordinates on gobans up to 19x19. _(Not so much inappropriate as it becomes a little unwieldy to write the moves when you have 2-digit numbers, which you would on boards 20x20 up. **There would be no problem in labelling the axes on the goban.**)_  

I would be very happy if I only had the option to use these coordinates on the 19x19 goban.

If it wasn't a difficult coding task then it would be good to have the option to use these coordinates on all goban sizes up to 19x19. Alternatively, it would be good to have the option on the 9x9 and 13x13 gobans. However 19x19 alone would be great!
9x9 goban: both axes are labelled 1 2 3 4 5 4' 3' 2' 1'
13x13 goban: both axes are labelled 1 2 3 4 5 6 7 6' 5' 4' 3' 2' 1'
20x20+ gobans: I suggest using normal numerals and dispensing with the "X" 

(Actually I prefer to use "*" rather than "'" when typing coordinates as it is easier to read; it stands out more, especially for those with less than ideal sight. Either would be fine when labelling the axes of the goban.)

ebifrier:
I added a shortcut to switch the display of coorinates, and supported for the 'Descriptive coodinates'.
'Descriptive coodinates' adopted '*' as described in the comments. Please check it out

GopherAl:
@ebifrier, it looks like you've already added descriptive coords. Fantastic! And much appreciated! I guess it'll be a while before it's available?

ebifrier:
If you can use git and npm, you can try it now.
This site is helpful for build & test.
[https://github.com/SabakiHQ/Sabaki/blob/master/docs/guides/building-tests.md](url)

yishn:
@GopherAl  Once merged, this feature will be available in the next release.

yishn:
I have some doubts about the nomenclature: The `japanese` coordinate system has no Japanese in it, so a better name would be simply `numbers`? 

Similarly, the `chinese` coordinate system has no Chinese in it, it just inverts the y axis and includes the letter I, which IMHO is just weird. Additionally, it might cause confusion, because we have coordinates detection in the comments. If you write `D4` in the comments it points to a completely different spot as indicated by the "Chinese" coordinate system.

At last, there's nothing `descriptive` or simple or natural about @GopherAl's coordinate system, I think a better name would be `relative`.

GopherAl:
@yishn @ebifrier I  agree that it is hard to come up with a good name! Relative, Descriptive and Natural all are appropriate, in my opinion. I think that Descriptive is the best of the three for the reasons below.

The notation is descriptive in the sense that you can read through an sgf and immediately visualise the relative positions of the moves. It is also obvious from the notation when a stone is played on a diagonal, for example. The position of any move is easily visualised. All points on diagonals have the form ii, ii', i'i or i'i'. Similarly, all points ij, ij', i'j, i'j', ji, ji', j'i, and j'i' are in the same position in relation to the corner in their quadrant of the board (taking symmetries into account). 

When discussing Joseki a similar notation is used and we commonly talk about the 3-3 pt, 4-4 pt, etc. Using this notation a Knight’s Approach to a 4-4 pt will be a variant of 2^6^ or 6^2^ where ^ indicates that there may or may not be the presence of a *. So 2^6^ and 6^2^ together summarise the 8 possible Knight’s Approaches to a 4-4 pt. Thus I feel it is an appropriate name.

For what it's worth it has similarities with the previously used chess notation, which was called Descriptive notation. See [(https://en.wikipedia.org/wiki/Descriptive_notation)] Columns were labelled accordingly to the starting positions of the pieces, and rows by their position looking from the White or Black side of the board, depending on who played the move. So White playing **P-K4** is equivalent to **e4** in the currently used Algebraic not'n.

yishn:
> The notation is descriptive in the sense that you can read through an sgf and immediately visualise the *relative* positions of the moves.

That's why it's "relative".

> When discussing Joseki a similar notation is used and we commonly talk about the 3-3 pt, 4-4 pt, etc. Using this notation a Knight’s Approach to a 4-4 pt will be a variant of 2^6^ or 6^2^ where ^ indicates that there may or may not be the presence of a *. So 2^6^ and 6^2^ together summarise the 8 possible Knight’s Approaches to a 4-4 pt. Thus I feel it is an appropriate name.

Again, I feel like this is an argument for calling it "relative". The fact that 2^6^and 6^2^ can summarize the knight's approaches to a 4^4^ is based upon the fact the the coordinate system is designed to be relative to the corners.

> For what it's worth it has similarities with the previously used chess notation, which was called Descriptive notation. See [(https://en.wikipedia.org/wiki/Descriptive_notation)] Columns were labelled accordingly to the starting positions of the pieces, and rows by their position looking from the White or Black side of the board, depending on who played the move. So White playing P-K4 is equivalent to e4 in the currently used Algebraic not'n.

I don't see any relation to the "descriptive notation" in chess here.

ebifrier:
One idea is to refer to Sensei's library.
[https://senseis.xmp.net/?Coordinates](https://senseis.xmp.net/?Coordinates)

For example:
  Global Coordinates -> 'Style A1 Coordinates' or 'A1 Style Coordinates'
  Japanese Coordinates -> 'Style 1-1 Coordinates' or '1-1 Style Coordinates'

'Chinese Coordinates' is difficult for me. If there is no demand, it may be removed.

yishn:
'A1 (Default)', '1-1', and 'Relative' work fine for me.

GopherAl:
Fine for me too.

ebifrier:
I renamed each coordinate system, and removed 'Chinese Coordinates'.

src/components/Goban.js
```diff
@@ -262,9 +262,34 @@ export default class Goban extends Component {
...
+
+    let relativeCoord = (x, size) => {
+      let halfSize = Math.ceil(size / 2)
+      let ix = size - x + 1
+      if (x <= halfSize && x < 10) return x.toString()
+      else if (ix < halfSize && ix < 10) return `${ix}*`
+      else return 'X'
+    }
+
+    let getCoordTransformer = coordinatesType => {
```
yishn: This variable might better be called `getCoordFunctions` since it might get confused with `transformCoords()`.

src/menu.js
```diff
@@ -554,10 +555,42 @@ exports.get = function(props = {}) {
...
+              label: i18n.t('menu.view', '&A1 (Default)'),
+              type: 'checkbox',
+              checked: !!showCoordinates && coordinatesType === 'A1',
+              click: () => {
+                setting.set('view.show_coordinates', true)
+                setting.set('view.coordinates_type', 'A1')
+              }
+            },
+            {
+              label: i18n.t('menu.view', '1-1'),
```
yishn: Is it not possible to put a keyboard hint on a number, like `&1-1`?

src/components/Goban.js
```diff
@@ -262,9 +262,34 @@ export default class Goban extends Component {
         board.height
       )
 
+    // Calculate coordinates
+
+    let relativeCoord = (x, size) => {
```
yishn: For fewer top-level variables and cleaner code, I suggest you move this function into the `else if (coordinatesType === 'relative')` branch.

ebifrier:
Ok, I refactored names and more.

GopherAl:
@ebifrier Great work! Relative coordinates are great. They would be perfect except for one thing. On the 19x19 goban **X** should be used in place of **10**. Would you please modify the code to do this. Once again, many thanks for what you (and @yishn of course) have already done.

GopherAl:
Sorry, there is actually an important defect other than the above. The axes should be labelled as in the diagram below. The coordinates in the bottom left quadrant are identical to the top right quadrant of the mathematical number plane. The negative coordinates (i.e. those starred **"*"**) should start from the right side counting towards the left and from the top counting downwards. 

People in general are accustomed to the mathematical number plane where numbers in the top right quadrant are all positive and increase to the right and upwards starting from the origin (bottom left). Similarly, 2-dimensional charts, graphs, etc, most commonly increase from an "origin" in the bottom left.

Labelling the bottom left quadrant of the goban from an "origin" in the bottom left is the natural way to do it which most (if not all go players) are already accustomed to seeing on the number plane and in charts etc. Labelling the axes as in the diagram below satisfies the mathematically intuitive rules:

- row numbers in the bottom half of the goban increase moving upwards towards row X
- column numbers increase when counting from the left towards the middle (column X)
- the numbers increase starting from the right or top and moving "backwards" towards the centre (but are labelled with a "*" to indicate that these numbers begin from either the top or right side of the goban.

**X** (capital X, the Roman numeral for 10) should be used on the **19x19 goban**. That greatly simplifies the process of writing down moves. For example, a move played at the central star point is written XX rather than 1010. Other examples are 3X versus 310 and X5* versus 105*.

image

yishn:
> People in general are accustomed to the mathematical number plane where numbers in the top right quadrant are all positive and increase to the right and upwards starting from the origin (bottom left). Similarly, 2-dimensional charts, graphs, etc, most commonly increase from an "origin" in the bottom left.

Due to the symmetry of the coordinate system, there are four "origins", therefore it really doesn't matter too much which corner is labeled with a *. But fine, we can change that if that makes you happy.

> X (capital X, the Roman numeral for 10) should be used on the 19x19 goban.

To use X instead of 10 doesn't make any sense for board sizes bigger than 19. Switching from arabic numbers to roman numbers when we get to 10 sounds terrible. To make just one exception for a 19x19 goban seems weird to me. Keep also in mind that we support rectangular boards.

GopherAl:
> Due to the symmetry of the coordinate system, there are four "origins", therefore it really doesn't matter too much which side is labeled with a *. Also, there are no negative numbers involved. But fine, we can change that if that makes you happy.

True, Sensei's library doesn't explicitly specify that the axes described have an origin at the bottom left. However it is implied, since that is the "normal" position of the "origin" in mathematics, charts, graphs, etc, and no other origin is specified. (SGF code is an exception, I believe. However that can hardly be said to be in common usage.) It also makes the best intuitive sense, as I outlined in detail in my previous comment. 

True, there are no negative _numbers_ involved. However the * indicates that you are counting in a negative (or opposite) _direction_ from the relevant side of the board. (Counting from the left or bottom may be viewed as moving in a positive direction. Then counting from the right or top may be seen as counting in a negative direction.)

It would make me **very happy** if you make that change!

> To use X instead of 10 doesn't make any sense for board sizes bigger than 19.

I agree totally. I didn't request that.

> To make just one exception for a 19x19 goban seems weird to me. 

True, it can be seen as weird. However it makes perfect sense on gobans up to and including 19x19. It ensures that all moves can be recorded in just two main characters plus one or two stars (*) if applicable. Also, this coordinate system is unlikely to be used on gobans larger than 19x19 anyway. (Probably no-one will ever notice the "weirdness" of X being replaced by 10 on larger gobans.)

I imagine Relative coordinates will mainly be used in analysis, etc, on the 19x19 goban. Relatively little analysis is done on gobans of other sizes. (I am aware only of discussions in books for beginners.) **I will be _ecstatic_ if you make this exception!**

> BTW, writing "39" to reference the point "3-9" can also cause misunderstandings, so a better way to write coordinates is always with a dash in between, like "3-10" or "10-5*".

In practice there is no confusion whatsoever on gobans up to 19x19 (especially when using X rather than 10). For larger gobans it would certainly be preferable to add a dash in between, as you suggest.

> BTW, do you have any real life usage of this coordinate system? Sensei's Library doesn't specifically say anything which side should get a *.

I and others use this system (including using X for 10) every day. The side that gets the * is implicit in the Sensei's Library description for all of the reasons that I outlined in my above post. It makes perfect sense to me and the other users that I am familiar with.

GopherAl:
@yishn Many thanks and **Stay Healthy!**

yishn:
I've implemented everything you wanted. But just to play the devil's advocate here:

> However the * indicates that you are counting in a negative (or opposite) direction from the relevant side of the board.

There's no side of the board that's more relevant than others. Although you can interpret the * indicator any way you want, interpreting it as negative numbers does not make sense: Why would we have a non-sequential jump from 10 to -9 all of a sudden in the middle of the board?

> In practice there is no confusion whatsoever on gobans up to 19x19 (especially when using X rather than 10). 

In practice there is no confusion on gobans up to 19x19 if you use 10 rather than X either.  "310" is unambiguously "3-10", since the alternative "31-0" is invalid.

>  It ensures that all moves can be recorded in just two main characters plus one or two stars (*) if applicable.

That means, the maximum length of a move would be four characters, which does not change if you use 10 over X.

> I and others use this system (including using X for 10) every day.

You do realize that answering a question about real life usage with "I and others use this system [...] every day" does not really answer the question.

GopherAl:
> I've implemented everything you wanted.

@yishn Many thanks!!! I've responded to most of your comments if you are interested.

> There's no side of the board that's more relevant than others. Although you can interpret the * indicator any way you want, interpreting it as negative numbers does not make sense: Why would we have a non-sequential jump from 10 to -9 all of a sudden in the middle of the board?

Here "relevant" refers to "left or bottom" when they are viewed as starting points for moving in a positive direction. (If you counted down from the top of the goban then the top would be the "relevant" side when talking about counting from the opposite side (bottom in this example) in the "negative" direction.)

As you have stated previously (when deciding to name the coordinates "Relative"), this coordinate system is totally based on the 4 corners of the goban. No matter what corner you start counting from, the numbers increase as you move towards the centre. The * simply indicates counting from the right if it is the 1st symbol of the pair denoting a move or it indicates counting downwards from the top if describing the 2nd symbol. So 4[star]7[star] describes the point that is 4 from the right and 7 from the top of the goban. (Clarification obviously required in Sensei's Library.)

So we are talking about **two opposing directions rather than positive or negative numbers**. The opposing directions meet in the middle of the goban and we have the transition you refer to. This is inherent in having a coordinate system that is relative to the corners. X is the 10th row or column counting from any direction; as you count past X you reverse direction. So moving from one side to the opposite side the numbers increase until you reach X then decrease from X/10 if you continue to head in the same direction. 

> In practice there is no confusion on gobans up to 19x19 if you use 10 rather than X either. "310" is unambiguously "3-10", since the alternative "31-0" is invalid.

True, but it is unwieldy and, dare I say, ugly! Using X keeps it clean and simple (in my eyes anyway).



    BTW, do you have any real life usage of this coordinate system? Sensei's Library doesn't specifically say anything which side should get a *.

> I and others use this system (including using X for 10) every day. The side that gets the * is implicit in the Sensei's Library description for all of the reasons that I outlined in my above post. It makes perfect sense to me and the other users that I am familiar with.

>   _I and others use this system (including using X for 10) every day._
You do realize that answering a question about real life usage with "I and others use this system [...] every day" does not really answer the question.

Re "real life usage", I can only speak from my own experience. Sensei's Library's description obviously needs clarification and expansion. There are many reasons for interpreting the description in the way that I have which I have endeavoured to explain.

Issue #649, opened 2020-03-14
Option to hide heatmap in analysis mode


yishn:
Sabaki should be able to record winrate graph without showing variations on the board.

SHKD13:
Would be nice to play with engines without seeing their variations, like it was in v0.43

PR #668, opened 2020-03-20
Add the option to switch on/off the display of variations

ebifrier:
This enables to switch on/off the variasions display started by mouse over.
It corresponds to the last comment in #649 .

### DIFF
--- a/src/components/ContentDisplay.js
+++ b/src/components/ContentDisplay.js
@@ -87,9 +87,7 @@ export default class ContentDisplay extends Component {
 
       sabaki.setState({playVariation: {sign, moves, sibling}})
 
-      if (setting.get('board.variation_instant_replay')) {
-        currentTarget.style.backgroundSize = '100% 100%'
-      } else {
+      if (setting.get('board.variation_replay_mode') === 'move_by_move') {
         clearInterval(this.variationIntervalId)
         this.variationIntervalId = setInterval(() => {
           if (counter >= moves.length) {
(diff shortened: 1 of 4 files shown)


yishn:
I think what @SKHD wanted was the ability hide the heatmap completely. I don't see any reason why we would disable playing out variations on mouse over. Just don't mouse over if you don't want to see the variation played out.

ebifrier:
What you say is theoretically correct. However, it is difficult for a careless person like me to always be careful not to hover over. I'm sure that this option is useful for careless people.

yishn:
Then I would prefer to merge this option into the Preferences drawer along with 'Instantly play out analysis variations on board'. We can have a select box instead with the options 'Don't Play', 'Instantly', and 'Move by Move' or something like that.

ebifrier:
Ok. That's a nice idea! I'll try to fix it.

SHKD13:
@yishn Not exactly. I don't wanna see any candidate moves on board during the game mode, just a clean board + winrate graph in a top right corner like it was in v0.43.
I see no reason to display candidate moves on board in any mode except special analysis mode F4.

yishn:
@SHKD13 Correct me if I'm wrong, but it sounds like you want to hide the heat map. Win rate data should only be recorded if you're in analysis mode anyway, otherwise there will be no analysis.

SHKD13:
@yishn Yes, would be nice to have an option to disable the heat map with candidate moves during the game mode.
Otherwise this is impossible to play with engine by yourself when it shows you all the best moves

ebifrier:
I added the new variable 'board.variation_replay_mode' to the setting, and removed   'board.variation_instant_replay'. Please review it.

Also, I set the title of the preference item to 'Replay mode of analysis variations:', it may be a bit weird. Please let me know if there is an alternative.

ebifrier:
Thank you for your reply! Thanks a lot!

Issue #586, opened 2019-06-22
Preserve window maximized state


ivysrono:
as title.

yishn:
We should probably save the window state along with window dimensions when closing.

mdm:
I'd like to work on this if nobody else is.

yishn:
I'm unassigning this issue since there are no updates.

ivysrono:
```
const {screen} = require('electron')
const display = screen.getPrimaryDisplay()
const dimensions = display.workAreaSize
let defaults = {
  'window.height': dimensions.height,
  'window.width': dimensions.width
}
```

`Error: The 'screen' module can't be used before the app 'ready' event`

ivysrono:
#670

yishn:
Please don't close this issue until a PR is merged.

PR #670, opened 2020-03-21
Preserve window maximized state

ivysrono:
#586 
This breaks package, at least portable will not work.
It needs `node_modules/{electron-store,conf,ajv,debounce-fn,mimic-fn,dot-prop,is-obj,env-paths,json-schema-typed,make-dir,semver,onetime,pkg-up,find-up,locate-path,p-locate,p-limit,path-exists,write-file-atomic,imurmurhash,is-typedarray,signal-exit,typedarray-to-buffer,type-fest,p-try,uri-js,……}/**/*",`
I'm not sure how much the `……` should be.
works: `"!node_modules"` => `"node_modules"`

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -46,11 +46,6 @@ function newWindow(path) {
     setting.set('window.maximized', false)
   })
 
-  window.on('resize', () => {
-    setting.set('window.width', window.getSize()[0])
-    setting.set('window.height', window.getSize()[1])
-  })
-
   window.on('closed', () => {
     window = null
   })
--- a/src/main.js
+++ b/src/main.js
@@ -13,12 +13,6 @@ function newWindow(path) {
       process.platform === 'linux' ? resolve(__dirname, '../logo.png') : null,
     title: app.name,
     useContentSize: true,
-    width: setting.get('window.preWidth')
-      ? setting.get('window.preWidth')
-      : setting.get('window.width'),
-    height: setting.get('window.preHeight')
-      ? setting.get('window.preHeight')
-      : setting.get('window.height'),
     width: setting.get('window.width'),
     height: setting.get('window.height'),
     minWidth: setting.get('window.minwidth'),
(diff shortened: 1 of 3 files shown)


yishn:
I don't see any reason why we have to install the `electron-store` dependency?

ivysrono:
> 
> 
> I don't see any reason why we have to install the `electron-store` dependency?

if no default value in settings.js, can not get next time?

yishn:
I don’t understand, can you elaborate?

ivysrono:
If I don't [add](https://github.com/SabakiHQ/Sabaki/pull/670/commits/dbf50fc26fcb634f94f84d4f8f13f22c3f2bc0aa#diff-c8842ba9921130e996b2439d6ae5f5c1R214) default values in `settings.js` , I get `null` after re-launch.
This is the reason I import electron-store at first.

yishn:
Yes, you need to add default values first to register them.

src/setting.js
```diff
@@ -210,7 +210,10 @@ let defaults = {
   'window.height': 604,
   'window.minheight': 440,
   'window.minwidth': 526,
-  'window.width': 564
+  'window.width': 564,
+  'window.max': false,
```
yishn: Rename to `window.maximized`.

src/setting.js
```diff
@@ -210,7 +210,10 @@ let defaults = {
...
+  'window.preWidth': undefined,
```
yishn: Do we really need extra settings for that? Can we not just use `window.width` and `window.height`?

src/main.js
```diff
@@ -33,6 +33,24 @@ function newWindow(path) {
...
+  window.on('maximize', () => {
+    setting.set('window.maximized', true)
+  })
+
+  window.on('unmaximize', () => {
+    setting.set('window.maximized', false)
+  })
+
+  window.on('resize', () => {
+    setting.set('window.width', window.getSize()[0])
```
yishn: We should already be setting these automatically in [App.js, line 72](https://github.com/SabakiHQ/Sabaki/blob/master/src/components/App.js#L72), so I don't think setting it here again is necessary.

ivysrono:
Thank you for your guidance, I learned a lot.

Issue #656, opened 2020-03-16
Local settings folder for portable app


ivysrono:
After we reinstall the system, or when we use Sabaki between several different computers, we want a file to record the settings modified by the user. The content of this file is equal to settings.json, but has higher permissions. LizGoban, Firefox and other open source projects provide this feature.

yishn:
Why not just use `settings.json`?

ivysrono:
First, it can be modified once and used everywhere;
Second, settings.json may be modified during major version updates, and a separate user configuration file can avoid version conflicts.

yishn:
`settings.json` can also be used everywhere. If there's need to modify settings during an upgrade, this would also have to affect your custom settings file.

ivysrono:
emm, settings.json is hard to store and sync. Could we custom it's path?

yishn:
How does Firefox support changing the settings path?

ivysrono:
[User.js file - MozillaZine Knowledge Base](http://kb.mozillazine.org/User.js_file)
[1.1 Overview · ghacksuserjs/ghacks-user.js Wiki](https://github.com/ghacksuserjs/ghacks-user.js/wiki/1.1-Overview)

There are more disadvantages of direct sync `settings.json`:
As # 586 has not been resolved, users may need to customize the width and height, but their settings may not apply to another computer.

LizGoban support `[config.json](https://github.com/kaorahi/lizgoban#to-configure-it)` to set sth, for example, engines, path, and so on.

I would like to put `config.json` into the same folder as `sabaki-**-portable.exe`, to get realy `portable`.

yishn:
Maybe for the portable version, say `path\to\sabaki-**-portable.exe`, we will resolve the settings folder as `path\to\sabaki-profile`, and for non-portable, it stays at `%APPDATA%\Sabaki`

ivysrono:
IMO, it is ok.

yishn:
Would you like to submit a PR? We set the setting folder path in [setting.js, line 7](https://github.com/SabakiHQ/Sabaki/blob/master/src/setting.js#L7).

ivysrono:
I'll try the next few days.

ivysrono:
> 
> 
> Would you like to submit a PR? We set the setting folder path in [setting.js, line 7](https://github.com/SabakiHQ/Sabaki/blob/master/src/setting.js#L7).

Why sabaki always use sync function?
Could we use async?
[...]

PR #671, opened 2020-03-22
Local settings folder for portable app

ivysrono:
#656

### DIFF
--- a/src/setting.js
+++ b/src/setting.js
@@ -3,8 +3,15 @@ const fs = require('fs')
 const path = require('path')
 const {app, BrowserWindow} = require('electron')
 
+const portableDir = process.env.PORTABLE_EXECUTABLE_DIR
+if (portableDir) {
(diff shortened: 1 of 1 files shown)


src/setting.js
```diff
@@ -3,8 +3,15 @@ const fs = require('fs')
...
+const portableDir = process.env.PORTABLE_EXECUTABLE_DIR
+if (portableDir) {
+  confirmPath = portableDir
+} else {
+  confirmPath = app.getPath('userData')
+}
+
 for (let dir of [
-  (exports.userDataDirectory = app.getPath('userData')),
+  (exports.userDataDirectory = confirmPath),
```
yishn: The variable `confirmPath` is not defined. We could also inline the logic here:

```suggestion
  (exports.userDataDirectory = process.env.PORTABLE_EXECUTABLE_DIR || app.getPath('userData')),
```

Issue #664, opened 2020-03-18
Reintroduce manage engines


benjaminvdb:
I really like that I can close all of Sabaki's panels and just focus on reviewing. Personally, I don't need to see the GTP output, so I'll just keep that closed. However, because the `Manage Engines...` menu item is now moved exclusively to the `Engines Sidebar`, I first need to open that, attach an engine and then close it again. I do like Sabaki's new way of dealing with engines, but this one thing is a bit inconvenient for me, and possibly for others.

One way around this is to reintroduce the `Manage Engines...` menu item to the `Engines` menu.

Another, even more convenient, solution would be to make it easier to start a review game mode. For example, if I would be able to choose a preferred engine (if there is more than one configured, otherwise, select the one engine as the default), and automatically start it whenever I hit `Toggle Analysis` (or the `F4` shortcut). Load game -> hit F4 -> start reviewing. Wonderful!

What do you think?

ebifrier:
I agree that it is a very nice feature. 

Technically, it is possible to initialize the engine and start the analysis at the same time.
The point to consider is that it may need to add new option to choose 'Analysis Engine' in the engine list.
It can be an option to use the engine at the top of the list, but changing the order of the list is not easy.

benjaminvdb:
> It can be an option to use the engine at the top of the list, but changing the order of the list is not easy.

I would be fine with a radio button that selects the default engine.

PR #673, opened 2020-03-22
Add default engine for analysis

ebifrier:
This enables to set the default engine for the analysis and start analysis without opening 'Engine Sidebar'. To achieve this, I added a new event called 'initialized' to EngineSyncer.
#664 

Note: the default engine is selected by checkbox instead of radiobutton, so the analysis starts with the top engine that is checked as the default engine. This allows you to avoid selecting a default engine and simplifies the code when adding or removing engines.

### DIFF
--- a/src/menu.js
+++ b/src/menu.js
@@ -16,6 +16,7 @@ exports.get = function(props = {}) {
   let selectTool = tool => (
     sabaki.setMode('edit'), sabaki.setState({selectedTool: tool})
   )
+  let analyzeEngineInitializing = false
 
   let {
     disableAll,
@@ -458,13 +459,32 @@ exports.get = function(props = {}) {
                 .map(syncer => syncer.id)[0]
 
             if (syncerId == null) {
-              dialog.showMessageBox(
-                i18n.t(
-                  'menu.engines',
-                  'None of the attached engines support analysis.'
-                ),
-                'info'
-              )
+              let defaultEngine = setting
+                .get('engines.list')
+                .filter(engine => engine.analysis)[0]
+
+              if (defaultEngine != null) {
+                if (!analyzeEngineInitializing) {
+                  analyzeEngineInitializing = true
+
+                  sabaki.attachEngines([defaultEngine], ({syncer, error}) => {
+                    if (error != null) {
+                      dialog.showMessageBox(error.message, 'error')
+                    } else {
+                      sabaki.startAnalysis(syncer.id)
+                    }
+                    analyzeEngineInitializing = false
+                  })
+                }
+              } else {
+                dialog.showMessageBox(
+                  i18n.t(
+                    'menu.engines',
+                    'None of the default engines for analysis.'
+                  ),
+                  'info'
+                )
+              }
               return
             }
 
--- a/src/menu.js
+++ b/src/menu.js
@@ -16,7 +16,6 @@ exports.get = function(props = {}) {
   let selectTool = tool => (
     sabaki.setMode('edit'), sabaki.setState({selectedTool: tool})
   )
-  let analyzeEngineInitializing = false
 
   let {
     disableAll,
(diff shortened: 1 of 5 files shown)


src/menu.js
```diff
@@ -458,13 +459,32 @@ exports.get = function(props = {}) {
...
+                .get('engines.list')
+                .filter(engine => engine.analysis)[0]
+
+              if (defaultEngine != null) {
+                if (!analyzeEngineInitializing) {
+                  analyzeEngineInitializing = true
+
+                  sabaki.attachEngines([defaultEngine], ({syncer, error}) => {
+                    if (error != null) {
+                      dialog.showMessageBox(error.message, 'error')
```
yishn: Unfortunately, `error.message` is untranslatable. Either show a message box with a generic error message, or we ignore errors as before. The user can check the GTP console for errors.

src/menu.js
```diff
@@ -458,13 +459,32 @@ exports.get = function(props = {}) {
...
-              )
+              let defaultEngine = setting
+                .get('engines.list')
+                .filter(engine => engine.analysis)[0]
+
+              if (defaultEngine != null) {
+                if (!analyzeEngineInitializing) {
+                  analyzeEngineInitializing = true
+
+                  sabaki.attachEngines([defaultEngine], ({syncer, error}) => {
```
yishn: I don't like the callback function here. `attachEngines` actually returns an array of `EngineSyncer`s that has been attached, so you can write something like:

~~~js
let [defaultEngineSyncer] = await sabaki.attachEngines([defaultEngine])

defaultEngineSyncer.controller.on('initialized', () => { /* ... */ })
~~~

src/menu.js
```diff
@@ -458,13 +459,32 @@ exports.get = function(props = {}) {
...
-                ),
-                'info'
-              )
+              let defaultEngine = setting
+                .get('engines.list')
+                .filter(engine => engine.analysis)[0]
+
+              if (defaultEngine != null) {
+                if (!analyzeEngineInitializing) {
+                  analyzeEngineInitializing = true
```
yishn: I don't like this global variable here in `menu.js`. We could consider blocking the whole UI with `sabaki.setBusy(true)` (then menu items will be disabled) and after the engine is initialized, call `sabaki.setBusy(false)`.

src/modules/enginesyncer.js
```diff
@@ -112,7 +112,9 @@ export default class EngineSyncer extends EventEmitter {
                 this.controller.sendCommand(Command.fromString(command))
               )
           : [])
-      ]).catch(noop)
+      ])
+        .then(() => this.controller.emit('initialized', {}))
```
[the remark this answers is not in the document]
ebifrier: Ok, I'll fix it.

src/menu.js
```diff
@@ -458,13 +459,32 @@ exports.get = function(props = {}) {
...
+                      sabaki.startAnalysis(syncer.id)
+                    }
+                    analyzeEngineInitializing = false
+                  })
+                }
+              } else {
+                dialog.showMessageBox(
+                  i18n.t(
+                    'menu.engines',
+                    'None of the default engines for analysis.'
```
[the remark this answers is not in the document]
ebifrier: Thanks

src/menu.js
```diff
@@ -458,13 +459,32 @@ exports.get = function(props = {}) {
...
-              dialog.showMessageBox(
-                i18n.t(
-                  'menu.engines',
-                  'None of the attached engines support analysis.'
-                ),
-                'info'
-              )
+              let defaultEngine = setting
+                .get('engines.list')
+                .filter(engine => engine.analysis)[0]
```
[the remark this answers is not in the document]
ebifrier: I'll fix it.

src/menu.js
ebifrier: I changed to display a generic error message.

src/menu.js
ebifrier: Ok, I removed 'analyzeEngineInitializing' and use sabaki.state.busy and sabaki.setBusy() instead.

src/menu.js
ebifrier: Thanks for your sample code! It is very helpful for me. Only one thing is that since the 'initialized' event may occur between sabaki.attachEngines() and defaultEngineSyncer.on('initialized', ...), I added 'autoStart' option to sabaki.attachEngines() to prevent it. However, it is very unlikely that this will happen, so I'll remove it if you don't need it.

src/menu.js
```diff
@@ -458,13 +459,32 @@ exports.get = function(props = {}) {
                 .map(syncer => syncer.id)[0]
```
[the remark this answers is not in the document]
ebifrier: I changed the priority of the analytics engine to use, but this algorithm is a bit more difficult for me. Please let me know if there is any difference from your intention.

ebifrier:
I've made changes to the code, so please let me know if anything is missing or modified. Can I ask for additional reviews?

yishn:
Yes, don't worry, I have this PR on my radar. Despite the pandemic, I'm a little bit busy this/next week.

ebifrier:
I'm sorry to hurry. Please take care of yourself!

yishn:
@ebifrier Would it be possible for you to reopen this PR? @apetresc Are you able to try this change out when reopened?

PR #698, opened 2020-04-17
Call spawn with absolute path if the file exists, or relative path if it doesn't

### LINKED ISSUE #695: Engines do not honor system paths on Linux
I use ssh to run my engines remotely on a more powerful machine.

For v0.43.3 and v0.50.0 I used "ssh" as my engine, works fine.
For v0.51.1 "ssh" doesn't work anymore, "/usr/bin/ssh" has to be specified.

Basically this is the same problem as pointed out in [issue 117](https://github.com/SabakiHQ/Sabaki/issues/117).

Sabaki version: v0.51.1
OS: Linux (SparkyLinux based on Debian testing branch, Kernel 5.6.0)

I think the issue has been introduced in [this commit](https://github.com/SabakiHQ/Sabaki/commit/f9bd1f30f63bd631064aff9f32f2d82a24491ac1).

ebifrier:
fix #695

### DIFF
--- a/src/modules/enginesyncer.js
+++ b/src/modules/enginesyncer.js
@@ -1,6 +1,6 @@
 import {remote} from 'electron'
-import fs from 'fs'
 import EventEmitter from 'events'
+import {existsSync} from 'fs'
 import {dirname, resolve} from 'path'
 import argvsplit from 'argv-split'
 import {v4 as uuid} from 'uuid'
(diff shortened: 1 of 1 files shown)


src/modules/enginesyncer.js
```diff
@@ -87,8 +88,19 @@ export default class EngineSyncer extends EventEmitter {
...
-    this.controller = new Controller(resolve(path), [...argvsplit(args)], {
-      cwd: dirname(resolve(path))
+    let fileExistsSync = path => {
+      try {
+        fs.accessSync(path, fs.F_OK)
+        return true
+      } catch {
+        return false
+      }
+    }
```
yishn, line 98: Why not simply use `fs.existsSync()`?

ebifrier: My understanding is that fs.existsSync is deprecated, but that information was outdated. I modified it to use fs.existsSync.

yishn:
Thanks a lot for your PR!

PR #700, opened 2020-04-20
Detach all engines when the window closed

ebifrier:
Running the engine over SSH may leave the engine's processes on the remote computer. This PR prevents it.
Also, if await sabaki.detachEngines() is done before sabaki.window.close(), it may take a long time to close the window, so await is at the end.

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -181,16 +181,15 @@ class App extends Component {
 
       evt.returnValue = ' '
 
-      setTimeout(async () => {
+      setTimeout(() => {
         if (sabaki.askForSave()) {
-          let promise = sabaki.detachEngines(
+          sabaki.detachEngines(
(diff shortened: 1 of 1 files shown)


src/components/App.js
```diff
@@ -181,11 +181,16 @@ class App extends Component {
...
+      setTimeout(async () => {
         if (sabaki.askForSave()) {
+          let promise = sabaki.detachEngines(
+            this.state.attachedEngineSyncers.map(syncer => syncer.id)
+          )
+
           gtplogger.close()
           this.closeWindow = true
           sabaki.window.close()
+          await promise
```
yishn, line 193: In this case, I don't think we even need to `await` the promise.

ebifrier: Oh, that's right. Thanks!

Issue #661, opened 2020-03-17
feature request: shortcut key for "show win rate" and "don't show"


lj739:

yishn:
Any suggestions?

PR #703, opened 2020-04-30
Make menu item keyboard shortcuts configurable in the settings

MarcelGruenauer:
Hello,

I forked Sabaki and made the menu item keyboard shortcuts configurable in the settings file.

I use Sabaki a lot for editing SGF files that contain lots of games, each of which contains quite some variations. So I'd like to have keyboard shortcuts for "Cut Variation", "Copy Variation" and "Paste Variation", among other things. But I understand that these shortcuts will not be useful for everyone, just like others would like to add or change shortcuts that are not useful for me.

Also I don't need shortcuts for rotating and flipping boards because that is not part of my usual workflow; I would rather have some of those shortcuts for other menu items.

So I thought it would be nice to be able to configure the shortcuts in the settings.json.

This is also related to issue #661 (feature request: shortcut key for "show win rate" and "don't show").

Thank you in advance for considering this pull request. If you have any suggestion, please let me know.


GoGameSpace:
I just noticed that the pull request has this line in settings.js:

    'shortcut.menu.play.toggle_player': 'CmdOrCtrl+Shift+1',

when the current version of the app has no such shortcut, so it should probably be

    'shortcut.menu.play.toggle_player': null,

I use this shortcut because I need this function a lot in my workflow.

yishn:
Thanks a lot for this PR. Here are my thoughts:

- I don't like all the keybindings be configured in `settings.json`. We should have a dedicated file for keybindings which only overwrites default keybindings, e.g. `keybindings.json`. This way, it will be easier to change default keybindings through updates while users still can retain their custom ones.
- I think it makes more sense to supply an `id` property for menu items in `menu.js`, (`accelerator` remains the default keybinding) and let the post-processing ([`processMenu`](https://github.com/SabakiHQ/Sabaki/blob/master/src/menu.js#L906)) resolve the actual keybinding from `keybindings.json`.
- Last but not least, I'm worried about keybinding conflicts. We should have a conflict check at startup and inform the user if there are any.

GoGameSpace:
I agree with your points. I've tried to make these changes but I'm not very familiar with the code yet.

There is a new branch in my fork, `topic-keybindings`; see https://github.com/GoGameSpace/Sabaki/tree/topic-keybindings

Could you please check whether this is more like what you had in mind?

So far there is no startup check for conflicting keybindings or invalid keybinding values. To do that, the default accelerators would need to be stored in a separate object as well. But I'm not sure how to go about that yet.

yishn:
Apologies for the late reply. You branch https://github.com/GoGameSpace/Sabaki/tree/topic-keybindings looks good for a first version. Please open update this PR, so it's possible to review it.

As for conflict check, maybe you can keep track of all the accelerators you assign to the menu items and check for duplicates that way.

MarcelGruenauer:
I'm not experienced with pull requests; isn't it open already? If not, what exactly should I do? Thanks.

yishn:
Do you mean this PR? This PR is trying to merge your `master` branch into Sabaki `master`. If you don't know how to update this PR, you can also just open a new PR merging your `topic-keybindings` branch into Sabaki `master`.

PR #715, opened 2020-05-08
Support large board size

### LINKED ISSUE #713: large board support
can remove the restriction of 25x25? to connect KataGo https://github.com/lightvector/KataGo/releases/tag/v1.3.5%2Bbs29

ebifrier:
fix  #713


yishn:
This isn't as easy as that. SGF can handle this fine, but coordinates won't display correctly above 'Z' and GTP engines stop functioning without us implementing KataGo's special syntax for higher coordinates.

Cyber-sport:
But can this be done as in q5go?

ebifrier:
That's for sure. Supporting KataGo's special syntax is going to take some work. Considering the number of users who use more than 25x25, it may be right not to support it.

PR #735, opened 2020-08-11
Add build for arm64

borongyuan:
Add build for arm64, so we can run Sabaki on Jetson Devices.
Tested on Jetson Xavier NX (JetPack 4.4).


borongyuan:
Thanks,
I'll try to port some engines to Jetson platform.

Issue #741, opened 2020-08-27
Entry in menu or shortcut to quit sabaki


rekstein:
Hi there,

First of all, thank you all for this great piece of software. Sabaki is neat and fantastic.

As I use a tiling window manager in GNU/Linux (i3wm) without UI border (ie. without a cross in an upper corner to close the window), I can't close Sabaki via a menu entry or a shortcut (ex. ctrl+w or alt+F4).
Thus I have to directly kill the Sabaki process. It's not really clean and convenient.
It would be great if quitting will be possible directly from Sabaki itself and not by killing its process or it's window.

Thank you by advance!
Regards,

yishn:
Does your tiling window manager not have a shortcut to close windows? Normally, <kbd>Alt+F4</kbd> closes a window as implemented by the OS.

rekstein:
@yishn 
Unfortunately none of the classic shortcuts (ie. ctrl+w ; ctrl+q ; ctrl+x ; ctrl+c ; ctrl+z ; alt+F4 etc.) work with Sabaki.
They work with other applications but not with Sabaki.
 when I list processes related to Sabaki I have :  
`
F S UID          PID    PPID  C PRI  NI ADDR SZ WCHAN  STIME TTY          TIME CMD
4 S myname      1481    1478  9  80   0 - 159623 -     16:58 ?        00:00:00 /tmp/.mount_sabakiFVtE3c/sabaki
1 S myname      1484       1 13  80   0 -  2903 -      16:58 ?        00:00:01 sabaki
4 S myname      1489    1481  0  80   0 - 48083 -      16:58 ?        00:00:00 /tmp/.mount_sabakiFVtE3c/sabaki --type=zygote
1 S myname      1491    1489  0  80   0 - 48083 -      16:58 ?        00:00:00 /tmp/.mount_sabakiFVtE3c/sabaki --type=zygote
0 S myname      1514    1481  2  80   0 - 84831 -      16:58 ?        00:00:00 /tmp/.mount_sabakiFVtE3c/sabaki --type=gpu-process --field-trial-handle=641201844416698950,1
0 S myname      1526    1481  1  80   0 - 63070 -      16:58 ?        00:00:00 /tmp/.mount_sabakiFVtE3c/sabaki --type=utility --field-trial-handle=641201844416698950,15696
[...]

PR #746, opened 2020-09-20
Added 'Quit' option under 'File'

HiggsTardigradeTau:
This adds a 'Quit' option under the 'File' menu. (#741)

### DIFF
--- a/src/menu.js
+++ b/src/menu.js
@@ -869,13 +869,15 @@ exports.get = function(props = {}) {
 
     appMenu.push(...items.slice(0, 2))
 
-    // Remove original 'Preferences' menu item
+    // Remove original 'Preferences' and 'Quit' menu items
 
     let fileMenu = findMenuItem('file')
     let preferenceItem = fileMenu.submenu.splice(
-      fileMenu.submenu.length - 2,
+      fileMenu.submenu.length - 4,
       2
     )[1]
+
+    fileMenu.submenu.splice(fileMenu.length - 2, 2)[1]
 
     appMenu.push(
       {type: 'separator'},
--- a/src/menu.js
+++ b/src/menu.js
@@ -109,7 +109,7 @@ exports.get = function(props = {}) {
         {type: 'separator'},
         {
           label: i18n.t('menu.file', '&Quit'),
-          accelerator: 'Alt+F4',
+          accelerator: 'CmdOrCtrl+Q',
           click: () => app.quit()
         }
       ]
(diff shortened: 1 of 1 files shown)


src/menu.js
```diff
@@ -105,6 +105,12 @@ exports.get = function(props = {}) {
           label: i18n.t('menu.file', '&Preferences…'),
           accelerator: 'CmdOrCtrl+,',
           click: () => sabaki.openDrawer('preferences')
+        },
+        {type: 'separator'},
+        {
+          label: i18n.t('menu.file', '&Quit'),
+          accelerator: 'Alt+F4',
```
yishn, line 112: We can't assign <kbd>Alt+F4</kbd> to quit, because on Windows <kbd>Alt+F4</kbd> is only supposed to close the current window not all windows of the application. <kbd>CmdOrCtrl+Q</kbd> is more appropriate.

Also, on macOS, the "Quit" menu item exists in [another menu already](https://github.com/SabakiHQ/Sabaki/blob/47a2693b15c6efaeed46a3e522cf6183dfbc0e68/src/menu.js#L898), so we need to hide these in macOS.

yishn (changes requested):

HiggsTardigradeTau:
Thank you! I will try to implement the necessary changes.

src/menu.js
HiggsTardigradeTau: I have changed the accelerator and hid the menu option on MacOS.

src/menu.js
```diff
@@ -863,14 +869,16 @@ exports.get = function(props = {}) {
...
 
     let fileMenu = findMenuItem('file')
     let preferenceItem = fileMenu.submenu.splice(
-      fileMenu.submenu.length - 2,
+      fileMenu.submenu.length - 4,
       2
     )[1]
 
+    fileMenu.submenu.splice(fileMenu.length - 2, 2)[1]
+
```
yishn, line 881: Instead you can write

~~~js
.splice(
  fileMenu.submenu.length - 4,
  4
)[1]
~~~

to remove the four last entries, so there's no need for two separate mutations.

yishn (changes requested):

src/menu.js
HiggsTardigradeTau: I have done so. I am sorry for requiring so much feedback, as I am new to JavaScript and I thought that it would be clearer if all that the code defining `preferenceItem`removed was the 'Preferences' option.

yishn (approved):
Thanks a lot!

rekstein:
Thank you all!

PR #752, opened 2020-10-31
Added accelerator for board.show_analysis to show/hide analysis results handly.

k-l-lambda:
Go learners can use analyzer as question & answer training. So switching analysis results shown/hidden is often to use.

I chose back quote (`) as the short key, that's can be discussed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,12 @@ All notable changes to this project will be documented in this file.
   (thanks to @ebifrier, #711)
 - Failed to load external links in comments (thanks to @baldor-f, #754)
 
+**Changed**
+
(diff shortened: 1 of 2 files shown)


yishn (changes requested):
I do not believe a backquote is a good accelerator. Maybe `CmdOrCtrl+Shift+H` instead.

apetresc:
As per @yishn's suggestion, I've changed the heatmap toggle to <kbd>CtrlOrCmd+H</kbd> and used <kbd>CtrlOrCmd+Shift+H</kbd> to _cycle_ through the different heatmap labels (which is a common thing to want to see).

apetresc (approved):

PR #753, opened 2020-10-31
A shortcut for game move by best point from analysis.

k-l-lambda:
I tried to use analyzer & AI player simultaneously, that's buggy when I back off and re-move frequently. So use an analyzer as an AI player is more practicable at present. And you can also control the pondering time flexibly.

I chose *Space* key as the shortcut for *move-by-analysis*, that's can be discussed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Sabaki develop](2020-10-31)
+
+**Added**
+
+- A shortcut for game move by best point from analysis.
+
 ## [Sabaki v0.51.1][v0.51.1] (2020-04-12)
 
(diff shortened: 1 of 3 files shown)


yishn:
Not sure why this is necessary? We have "Generate Move" which will let the engine choose the move for you.

k-l-lambda:
As I said, AI player synchronization is buggy and troublesome when changing game tree position frequently.

It's useful for me. Never mind.

yishn:
Would be nice if you could create an issue on how it is buggy with detailed reproduction steps and visualization.

k-l-lambda:
Try to keep turning mouse wheel when AI player is thinking.

It seems caused by synchronous GTP player communication. Meanwhile, the AI analyser is more flexible.

Issue #719, opened 2020-06-02
Crash on “View Comments” command when window width < screen resolution


upsided:
Resizing the window to a width < screen resolution, then selecting “View Comments” crashes Sabaki. This makes the interface unresponsive to any mouse clicks (though I can still hear stone sounds…). 

Closing the window and creating a new one  returns interaction to normal, with the comments sidebar now visible. However, choosing “view comments” again to hide the comment window  crashes this new window once more.

This crash happens in both cases: when you hide the comments sidebar, and when you try to show it.

Note: it is the width of the application (OS) window that seems to cause this problem, not the width of the “main” div.

I have tested at different screen resolutions, and as long as Sabaki consumes the entire width of the monitor, this crash doesn’t happen for me.

Mac OS High Sierra 10.13.6

This is this error:

index.html:1 
Uncaught (in promise) Error: Could not call remote method 'setContentSize'. Check that the method signature is correct. Underlying error: Error processing argument at index 0, conversion failure from 
Underlying stack: TypeError: Error processing argument at index 0, conversion failure from 
    at electron/js2c/browser_init.js:6732:67
    at IpcMainImpl.<anonymous> (electron/js2c/browser_init.js:6597:27)
    at IpcMainImpl.emit (events.js:210:5)
    at WebContents.<anonymous> (electron/js2c/browser_init.js:3873:23)
    at WebContents.emit (events.js:210:5)

    at electron/js2c/browser_init.js:6735:21
    at IpcMainImpl.<anonymous> (electron/js2c/browser_init.js:6597:27)
    at IpcMainImpl.emit (events.js:210:5)
    at WebContents.<anonymous> (electron/js2c/browser_init.js:3873:23)
    at WebContents.emit (events.js:210:5)

Promise.then (async)

_
@
bundle.js:1

b.setState
@
bundle.js:1

(anonymous)
@
bundle.js:68

emit
@
events.js:210

setState
@
bundle.js:1

g
@
bundle.js:6

n.click.listeners.<computed>
@
bundle.js:68

emit
@
events.js:210

onMessage
@
init.ts:50

apetresc:
@upsided This should be resolved on `master` now thanks to @cravay's patch.

If you have the ability to test the master branch, give it a show and let me know if the fix also works on your macOS environment (I believe it should). Otherwise, it will be in the v0.52.0 release soon.

PR #771, opened 2021-01-24
Fix crash on “View Comments” (#719)

cravay:
I also had issue #719 on my Windows 10 PC. `window.setContentSize` expects integer arguments but `widthDiff` is not (always) a whole number. I have verified that this fix works by building and running Sabaki on my PC.

I love Sabaki! Thanks a lot for all your hard work!


apetresc:
Thank you! :)

PR #784, opened 2021-04-10
Fix layout of edit mode when comment box is hidden

FanJeldmann:
When the comment box is hidden and edit mode is being turned on, the layout and therefore the mouse event handling on the game graph gets messed up (offset by some amount).

I fixed that.

Be aware that I am not very experienced with react, so I don't know if this is the best way of doing this.

Cheers

### DIFF
--- a/src/components/sidebars/CommentBox.js
+++ b/src/components/sidebars/CommentBox.js
@@ -280,8 +280,8 @@ export default class CommentBox extends Component {
     }
   }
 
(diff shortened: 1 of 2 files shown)


apetresc:
Thank you so much for diving in with this! After a bunch of investigation, I can't really find a cleaner way of accomplishing this either - so thank you so much for this implementation :)

Issue #487, opened 2018-12-20
Ability to automatically store LZ variations in sgf


vascopimenta:
Apologies if this is the wrong place to suggest features.

After some trial and error, I've finally been able to have Leela Zero working with Sabaki on macOS. Kudos to the developers for all the work put into supporting this.

As it stands, if one wants to save LZ suggested variations, they have to be manually added (right-click on a given spot on the heatmap -> Add variation). The resulting file can then be saved, and these variations/branches show up when one opens the sgf file. Also, the winrate graph is retained, even though at this stage it seems buggy.

I believe it would be a very interesting feature (especially to those of us relying on CPU-only versions of LZ) to be able to automatically have LZ attached and perform a full analysis of the game, let's say, providing a maximum thinking time per move, and have the n highest visited locations saved as branches for each position. Another interesting option would be to store those branches only if the variation in estimated winrate after a given move varied for more than x%. Ideally, also the heatmap, winrate %, number of visits would be saved and shown as well.

This would basically give you an sgf file with the analysis of your game, showing where you and your opponent made the most important mistakes, and would IMHO be a very useful tool for improving at Go. I imagine these features being implemented in a different mode/dialog box - say, Tools -> Game Analysis, where you would select engine to attach, thinking time, the above described criteria, etc., and then you would press start and wait for it to run over the whole game and save the output over the existing sgf file (or a new one, as an option).

Sorry for the not very focused suggestion, I hope the central ideas come across. Feel free to ask for clarifications, if you find the idea interesting, and thank you for Sabaki.

yishn:
Hi, thanks for the suggestion. FYI there's currently a separate tool for doing just that, called [GoReviewPartner](http://yuntingdian.com/goreviewpartner/) by @pnprog. You can feed it any sgf file and it will create an `rsgf` file which you can open with Sabaki. 

Doing a full game analysis will take a looong time, during which you cannot edit the game tree. You might as well use an external tool for that. Nevertheless, I'm considering such a feature for the future for Sabaki as well.

> Also, the winrate graph is retained, even though at this stage it seems buggy.

Please open a separate issue describing why it's buggy.

vascopimenta:
> Please open a separate issue describing why it's buggy.

Will do. Will try to find out first whether I can find reproducible steps to provoke the issue (suddenly the red bars disappear, and only the white graph remains visible).

vascopimenta:
[...]

PR #808, opened 2021-08-07
Support WINRATE property

kevinsung:
Part of #487 .

This lets Sabaki recognize when an SGF file has the non-standard WINRATE property attached to move nodes, in which case it would display a winrate graph even with no engine analysis session running. A winrate value from an engine analysis session would override the value from the SGF file.

Here I assume WINRATE represents the probability of black winning as afloat in [0, 1], but I can change this.

### DIFF
--- a/src/modules/sabaki.js
+++ b/src/modules/sabaki.js
@@ -196,7 +196,11 @@ class Sabaki extends EventEmitter {
       get winrateData() {
         return [
           ...this.gameTree.listCurrentNodes(state.gameCurrents[state.gameIndex])
-        ].map(x => x.data.SBKV && x.data.SBKV[0])
+        ].map(
+          x =>
+            (x.data.SBKV && x.data.SBKV[0]) ||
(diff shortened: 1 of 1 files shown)


apetresc:
Are these the same non-standard property names that things like [goreviewpartner](https://github.com/pnprog/goreviewpartner) emits in its `.rsgf` files? If we're going to adopt non-standard extensions, they should at least be _de facto_ standards.

Nevertheless, your contributions are very much appreciated @kevinsung :)

kevinsung:
@apetresc I have not tried goreviewpartner or any other tool that adds properties to SGF files, though I did read https://github.com/SabakiHQ/Sabaki/issues/487#issuecomment-450712215 by the main developer of goreviewpartner. It wasn't clear to me what the "BWWR" property mentioned there meant, so I didn't copy it. I agree that we should try to stick to de facto standards, so I understand that more discussion on these standards may be needed before merging this PR. Perhaps we should continue the discussion over at the linked issue.

kevinsung:
Closing due to resistance against non-standard SGF properties.

PR #809, opened 2021-08-07
Add heatmap from pre-analyzed moves

kevinsung:
Part of #487 .

This lets Sabaki recognize the non-standard SGF properties VISITS, WINRATE, and SCORELEAD in order to draw a heatmap even with no engine analysis session running. Values from a running engine analysis session would override the values from the SGF.

I assumed the following definitions for these properties, but I can change them:
- VISITS: number of visits
- WINRATE: probability of black winning as a float in [0, 1]
- SCORELEAD: number of points black is winning by

The diff is deceptively large because I indented an existing block by an additional level.

I have written a script that uses KataGo to add these properties to an SGF file: https://github.com/kevinsung/katago-analyze-sgf.

Screenshot:
Screenshot from 2021-08-07 14-57-46

### DIFF
--- a/src/components/Goban.js
+++ b/src/components/Goban.js
@@ -220,6 +220,7 @@ export default class Goban extends Component {
       paintMap = [],
       analysis,
       analysisType,
+      showAnalysis,
       highlightVertices = [],
       dimmedStones = [],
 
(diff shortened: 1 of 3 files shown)


yishn:
I don't like the non-standard SGF properties. Is anyone else using these? 🤔 I did intend for the heat map to be a fleeting (temporary) board layer as opposed to persisted in the SGF file. But more people do full-game analysis beforehand and want to look at the results. Maybe sync with @pnprog.

kevinsung:
Yeah, for better or worse, using non-standard SGF properties is the most straightforward way to support full-game analysis. I did reach out to @pnprog at #487 but he hasn't responded. I would also love to know if other tools use similar non-standard properties; I couldn't find any besides @pnprog's Go Review Partner.

kevinsung:
Closing due to resistance against non-standard SGF properties.

PR #818, opened 2021-08-21
Links to Antique and Jade themes.

billhails:
A couple more themes for you, hope you like them.

- [Antique](https://github.com/billhails/SabakiThemes/blob/main/antique/)
- [Jade](https://github.com/billhails/SabakiThemes/blob/main/jade/)

 I'll probably not do any more now.

There is another "semi-precious" theme in my repo but having finished it I don't really like it so I won't promote it here. I'll leave it as an easter egg if anyone goes looking.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,14 @@
 
 All notable changes to this project will be documented in this file.
 
-## [Sabaki v0.52.0][v0.52.0]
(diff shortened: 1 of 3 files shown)


billhails:
https://github.com/billhails/SabakiThemes/tree/main/blender - enjoy!

PR #823, opened 2021-09-07
Add score lead line to winrate graph

zogwarg:
Early PR at this stage, would welcome some suggestions to make cleaner. and for how to do the i18n correctly.

Content of this PR:
- When analyzing with katago,  black's score lead is added to a node property `SBKS` <-- Welcoming a better suggestion for the node property name
- Adds an toggleable option to display the score lead graph `Show Score Lead Graph`
- Adds a blue line representing score lead as an overlay to winrate graph
- Adds score lead in tooltip

Example image:
Screen Shot 2021-09-08 at 1 06 16

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -346,6 +346,7 @@ class App extends Component {
         showNextMoves: state.showNextMoves,
         showSiblings: state.showSiblings,
         showWinrateGraph: state.showWinrateGraph,
+        showScoreLeadGraph: state.showScoreLeadGraph,
         showGameGraph: state.showGameGraph,
         showCommentBox: state.showCommentBox,
         showLeftSidebar: state.showLeftSidebar,
--- a/src/components/Sidebar.js
+++ b/src/components/Sidebar.js
@@ -106,13 +106,15 @@ export default class Sidebar extends Component {
       treePosition,
 
       showWinrateGraph,
+      showScoreLeadGraph,
       showGameGraph,
       showCommentBox,
 
       graphGridSize,
       graphNodeSize,
 
-      winrateData
+      winrateData,
+      scoreLeadData
     },
     {winrateGraphHeight, sidebarSplit}
   ) {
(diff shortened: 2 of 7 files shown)


zogwarg:
Apparently coressponds quite well to https://github.com/SabakiHQ/Sabaki/issues/755

apetresc:
Thank you for beginning work on this :)

Is this ready for review? I can't seem to get it to work - when I toggle the "Show Score Lead Graph", it just toggles the game tree. I haven't had time yet to try to diagnose anything, just my first attempt.

zogwarg:
Hi,

It's a pleasure to contribute!
Are there big flagrant issues, or should there just be small corrections and esthetic changes ?

kevinsung:
I've tested this and found some issues:
- When the `Show Winrate Graph` option is unchecked, the entire graph disappears. I think the correct behavior should be to show the graph if either `Show Winrate Graph` or `Show Score Lead Graph` is checked. When only one of them is checked, only the corresponding data should be shown (the current behavior is correct when only winrate is checked but incorrect when only score lead is checked).
- The current shade of blue is a bit difficult to distinguish from the background. I think the color should be made lighter. I suggest using `#36f` and `#36d`.

kevinsung:
It's possible this is outside the scope of this PR, but the other features related to the win rate should be replicated for the score lead, namely, the displays in the graph showing the numerical value, as well as the difference bars in the graph. For the difference bars it might make sense to have an option to choose whether the bars correspond to win rate or score lead, and even to hide them altogether.

kevinsung:
Actually, having all information simultaneously displayed might be too cluttered. Perhaps the graph should have information corresponding to only one of win rate or score lead at any given time, with an option to toggle between them. Then the same colors could be used for the lines and bars in either scenario.

kevinsung:
This feature is a high priority for me so I went ahead and fixed the issues myself in #828 .

PR #828, opened 2021-10-11
Support displaying score lead in analysis graph

### LINKED ISSUE #755: scoreLead display on graph
Even when the scoreLead display is chosen, the right-hand side graph displays winrate only. I propose changing it so that scoreLead is displayed instead (or in addition).

I have a working solution, which I'd like to polish and send a PR, if you like the idea. I imagine the following steps:

- [ ] Adding an extra property to each move (SBKV is used for winrate, we could add SBKS?)
- [ ] Passing both winrate and scoreLead data to the winrate graph
- [ ] Adding some Y-axis labels (say 50% and 100% for winrate, and auto-computed values for scoreLead). Should we display both values on the graph (using left/right Y-axis scale), or just the one currently shown on the board? I'm leaning towards displaying just one - it's cleaner and more coherent.

kevinsung:
Fixes #755 . Closes #823 .

This PR is an alternative to #823. I did explain the issues with that PR in the comments there, but since this feature is a high priority for me I went ahead and implemented it myself. In contrast to #823, which adds a score lead line on top of the existing graph, this PR makes the graph show either win rate or score lead, but not both, with the same style in either case. The metric (winrate or score lead) shown on the graph always matches the metric shown in the heatmap.

There are many variables in the code that use the term "winrate" which is no longer accurate after this change, since it can mean either "winrate" or "score lead". To keep the diff small, I did not change those variable names in this PR. Those variables should probably be renamed in a future PR.

Screenshots (second image taken after clicking "Score Lead" in first image):
Screenshot from 2021-10-11 11-58-20
Screenshot from 2021-10-11 11-58-24

### DIFF
--- a/e2e/analysis-graph.spec.js
+++ b/e2e/analysis-graph.spec.js
@@ -120,4 +120,19 @@ test.describe('Analysis Graph', () => {
     // A score lead of 0 is a real, common value; the marker must still render.
     expect(await markerCount(page)).toBe(1)
   })
+
+  test('updates the data line when the metric is toggled at an unanalysed node', async ({
+    page,
+  }) => {
+    await goToStep(page, 0) // root: no SBKV/SBKS → data[currentIndex] == null
+    const winratePath = await dataLinePath(page)
+    expect(winratePath).toBeTruthy()
+
+    await setMetric(page, 'scoreLead')
+    const scoreLeadPath = await dataLinePath(page)
+
+    // data[currentIndex] is null for both metrics here, so the re-render must
+    // be driven by the analysisType change, not the current value.
+    expect(scoreLeadPath).not.toBe(winratePath)
+  })
 })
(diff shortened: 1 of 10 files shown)


petermao:
This is a nice, clean implementation of adding a score lead graph.  The keybinding (C-H) make it easy to toggle between the two modes.  I have tested out commit [6f0751a], but I like it well enough that locally, I'm running a merge of Kevin's `dev` branch and my own development branch.

kevinsung:
@apetresc @yishn Thanks so much for maintaining this wonderful project. What are my prospects for getting this PR merged?

apetresc:
Just a heads-up that I haven't ignored this PR either, it's just that this one requires much more thorough tests to be written (since currently Sabaki's integration/e2e tests cover engine analysis only shallowly). So I'm taking a detour through a different branch to beef up the test coverage there, and then use that to implement e2e tests for this one as well.

So, working on this one too, should be ready maybe in about 12-24 hours.

apetresc:
Rebased against master and the new engine integration test suite that this warranted. @kevinsung can you please do a round of manual testing and make sure I haven't broken anything? You are more familiar with this functionality than I am :)

kevinsung:
> Rebased against master and the new engine integration test suite that this warranted. @kevinsung can you please do a round of manual testing and make sure I haven't broken anything? You are more familiar with this functionality than I am :)

I just tested it and it still works. Thanks for working on this!

Issue #832, opened 2021-11-02
Minimal language support for speller in comments


zorgnac:
Sabaki v0.52.0 - linux x64

When editing comments, Sabaki underlines misspelled words with a red wave. The notion of misspelling is built-in according to some reference language - namely English - and is not affected by user interface localization. This makes editing quite uncomfortable, as you may type perfect spelling in your own language - or even worse use proper go vocabulary - and get distracted by those red waves you cannot do anything about.

Sabaki may, as a minimal support, either let the user choose not to underline misspelled, or to specify the list of languages he is editing for.

If anyone is interested, I've made patch that implements a demo of the latter alternative. It defines a new entry `comments.langs` in settings, and adjusts `setSpellCheckerLanguages` in `main.js:newWindow`. It's worth about 5 lines of code. There no UI here, it relies on manual editing of  `settings.json`.

apetresc:
That's a great point, thank you! Yeah, feel free to post up your initial pull request as a starting point :)

zorgnac:
Not sure I've done it the expected way.

However, here is pull request #833

zorgnac:
I've worked on this and added some UI around, namely a context menu in the comment box with option to replace misspelled or add to dictionary.

Shall I close the PR and replace by another one?

zorgnac:
![custom](https://user-images.githubusercontent.com/93589069/148386658-a06f90b1-33d1-4989-8f81-9786d13ce475.png)

PR #833, opened 2021-11-12
Spell Checker: minimal support for multiple languages (via settings)

zorgnac:
Demo code for issue #832


zorgnac:
Found some info about [spell checker in electron](https://www.electronjs.org/docs/latest/tutorial/spellchecker/)

zorgnac:
Exploring a little further, I've patched my copy of Sabaki to:
* add a spelling contextual menu, with possible addition to `Custom Dictionary`
* define a default set of Romaji go terms in `Custom Dictionary`

I've found the contextual menu somehow tricky; it's a more in-depth patch, where I'm absolutely convinced I'm out of coding standards.

You've got a demo (together with other customization) in my [focal](https://github.com/zorgnac/Sabaki/tree/focal) branch.

apetresc:
Thanks for putting this together, and apologies it sat so long. Closing this one, but I wanted to explain why rather than just letting it lapse.

On current Electron the built-in spellchecker already follows your OS locale automatically: on macOS it uses the native system spellchecker (which handles multiple languages on its own), and on Windows/Linux it auto-populates from your OS locale on launch. So spellchecking in your system's language works out of the box now, with no config.

The one thing that isn't covered is spellchecking in a language different from your OS locale (or several at once on Windows/Linux). That's a real gap, but it's niche enough that I don't think it justifies the settings surface right now, especially since the workaround is to adjust your OS locale or add one. If more people come forward wanting multi-language spell checking in comments, I'm happy to revisit it as a small focused change against the current spellchecker API.

This PR has also drifted a long way from master over the years, so even the relevant parts would need a fresh start. Thanks again for the exploration, and for the pointers to the Electron spellchecker docs.

PR #834, opened 2021-11-13
Issue 773 - cycle through downstream variations

petermao:
For your consideration, here is an implementation of [Issue 773](https://github.com/SabakiHQ/Sabaki/issues/773).

Starting from the present node, this finds downstream forks and cycles through the possible variations by altering the `gameCurrents` array.  For myself, I prefer to have these new function calls bound to `Left` and `Right`, but in the interest of not surprising anyone, I bound these to `Shift+Left` and `Shift+Right`.

The function name is a bit jarring.  It doesn't act like a `goTo...`, so I didn't name it as such.

### DIFF
--- a/src/menu.js
+++ b/src/menu.js
@@ -411,12 +411,12 @@ exports.get = function(props = {}) {
           click: () => sabaki.goToSiblingVariation(1)
         },
         {
-          label: i18n.t('menu.navigation', 'Decrement Downstream &Variation'),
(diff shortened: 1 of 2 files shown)


apetresc:
I agree the name is a bit jarring but I couldn't find a better alternative either. Thanks for this, I think it's quite useful! :)

(Better late than never right? 😅)

Issue #848, opened 2022-01-04
Easy access to engine role (feature/enhancement)


zorgnac:
Sabaki v0.52.0 - linux/ubuntu x64

Here is some return of experience when using engines in Sabaki. I split here in three distinct topics, but it looks like it's about the same thing. There is no real issue here, just comments.

## Layout

While analyzing a game segment, I'm often changing the fact the engine replies or not to my moves: sometimes I want it play white, sometimes black, and sometimes not to play at all.

But opening and browsing the engine context menu to adjust its role is not very user friendly. I can feel it in particular when I receive an answer from the engine while I've just decided to explore a branch by myself. This is usually after an exhausting while doubtful mental reading: "No, don't do that! Where is it again? The `Engine` menu?"

I'm hoping it would be much more convenient if an engine in left sidebar always displays the two (black and white) stones, reflects its role by some style (e.g. opacity), and responds to a single click to toggle it. Currently, the engine only shows the color it's playing for.

## Startup

I've noticed that my first action when I open Sabaki on a game is almost always to load `katago`. But actually, I'm opening a **game** and I'm already thinking about it, I'm not opening `katago`.

Consequently, I'd like to configure some of my engines to load at startup (may say autoload the engine).

## Default role

My second action is to affect `katago` as analyzer.

You've already guessed: I'd like to configure the engine so that it takes some default role when it loads: black, white or analyzer.

The third point is tightly related to the second, but is logically distinct. And the easiness of access exposed in the first point would cope with a not-so-systematic default role. That's why I post all at once. What do you think?

sente361:
I find pressing the F4 button does most or all of what you are trying to achieve. I am using Windows but I presume that linux versions offer the same options?

1. Attach an engine. (The first image shows nothing ticked when you right-click on the engine.)
2. "Toggle Analysis" on or off as required. (See the second image.)

![image](https://user-images.githubusercontent.com/47664082/148162567-b76f8261-2aed-4dc9-98c2-e4c08a06e119.png)

![image](https://user-images.githubusercontent.com/47664082/148162360-180c8c42-0a37-4008-98fc-cef0d5d10c45.png)

zorgnac:
Well, yes I currently can Start/Stop analysis via `F4`. 

But when it's off, I don't have the heatmap anymore. I activate the heatmap with the menu `View/Show Heatmap`.

And when it's on, I'm not so comfortable with the noise my GPU produces and the fact the heatmap refreshes periodically.

[...]

PR #849, opened 2022-01-04
Engine layout

zorgnac:
Demo of layout part of #848

This keeps engine player icon visible, and makes them clickable to toggle player role. Displaying whether an engine plays black or white (or none of them) is controlled by CSS (here opacity). 
Setting CSS to `display: none` would restore the old layout, though icons are still clickable.


zorgnac:
Removing

PR #850, opened 2022-01-04
Autoload engine and give it a default role

zorgnac:
Demo of autoload/role in #848

* new field `boot` in engine's properties (array of symbols)
* `enginesyncer` emits `commands` when engine has answered to `list_commands`
* at startup, sabaki attaches engines that have `autoload` among `boot` symbols
* when receiving `commands` sabaki affects engine roles according to symbols `white`, `black`, or `analyze` in `boot`


zorgnac:
OK. removing it

PR #883, opened 2022-08-13
Upgrade to Shudan 1.7.x

### LINKED ISSUE #882: Adopt Shudan 1.7.0
There's some cool visual enhancements in Shudan 1.7.0 to be adopted.

apetresc:
This required an upgrade from Webpack 4 to Webpack 5 because Shudan 1.7.0 makes use of the nullish coalescing operator '??' which Webpack 4 fails to parse. Luckily, this was a pretty painless migration.

Closes #882.

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,8 +122,8 @@
     "prettier": "1.19.1",
     "react": "^16.13.1",
     "tmp": "^0.1.0",
-    "webpack": "^4.42.1",
-    "webpack-cli": "^3.3.11"
+    "webpack": "^5.74.0",
+    "webpack-cli": "^4.10.0"
   },
   "scripts": {
     "test": "mocha --require esm",
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "@sabaki/immutable-gametree": "^1.9.4",
     "@sabaki/influence": "^1.2.2",
     "@sabaki/sgf": "^3.4.7",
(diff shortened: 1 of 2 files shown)


apetresc:
Hmm, so `webpack --mode production` was working fine with Shudan 1.7.0 last night, but something in Shudan 1.7.1 this morning has broken it again. Needs investigation.

apetresc:
@yishn Do you have any insight here? It seems there's some sort of incompatibility with how imports are resolved. Ever since SabakiHQ/Shudan@966c924b4629261667e5710f105e5d46ac949d81 Sabaki is translating's Shudan's `require("preact/hooks")` into a search path of `[..]/node_modules/preact/dist/preact.min.js/hooks[.js|.json|.wasm]` which clearly seems wrong. I'm not fluent enough in all the new Node/ES/Webpack churn to make sense of which component's fault this is, do you have an idea off the top of your head?

yishn:
I have removed the "preact" alias in `webpack.config.js` and it seems to be working now

apetresc:
Indeed, not sure how I missed that :) Though I wish I understood what it was about `"type": "module"` (presumably) that changes the way the import path is generated.

Thanks @yishn!

yishn:
I think the reason was the introduction of a `preact/hooks` Import that did not exist before 😅

PR #941, opened 2023-07-02
Allows users to edit influence boxes when in Estimate/Scoring mode.

RobertChrist:
When in scoring or estimation mode, we now track any vertexes that the user clicks
via the state.estimateOverrides property.  In our app render method, we then
override the area influence map with these clicks.  Clicks cycle through black/white/
empty.  A +1, -1 is used in the equations to reflect the fact that black/white/empty values
in the influence map are tracked via values of [-1,0,1].

### DIFF
--- a/e2e/scoring-overrides.spec.js
+++ b/e2e/scoring-overrides.spec.js
@@ -1,0 +1,90 @@
+const {expect} = require('@playwright/test')
+const {test} = require('./fixtures/electron-app')
+const {loadSgfStringAndWait, waitForRender} = require('./helpers')
+
+// Manual territory overrides in scoring mode (state.estimateOverrides): clicking
+// an empty point cycles its area value one step per click, and toggling any
+// stone's life/death clears all overrides so a stale one can't drift against the
+// recomputed area map (the bug from the PR's 2024 review). Assertions check both
+// the state map and the rendered Shudan paint, so a broken render path can't hide
+// behind correct state.
+
+const SGF = '(;GM[1]FF[4]SZ[9];B[cc];W[gg])'
+
+function overrides(page) {
+  return page.evaluate(() => window.__sabaki.state.estimateOverrides)
+}
+
+// The territory paint Shudan renders for a vertex: 1 (black), -1 (white), or 0
+// (none). Scoring mode feeds the areaMap in as the paintMap, so this reflects
+// the override applied on top of the estimate.
+function paintAt(page, [x, y]) {
+  return page.evaluate(
+    ([vx, vy]) => {
+      let el = document.querySelector(
+        `.shudan-vertex[data-x="${vx}"][data-y="${vy}"]`,
+      )
+      if (el == null) return null
+      if (el.classList.contains('shudan-paint_1')) return 1
+      if (el.classList.contains('shudan-paint_-1')) return -1
+      return 0
+    },
+    [x, y],
+  )
+}
+
+async function click(page, vertex) {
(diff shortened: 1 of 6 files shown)


apetresc:
Similar apology here about the length of time it's taken me to get to it :)

This is cool, and seems to work well. But I think we need to also update the subroutine for marking stones alive/dead, since the changed values in the areaMap are clashing. See the video below for a demonstration of the issue:

https://github.com/SabakiHQ/Sabaki/assets/14872/bfcf2a96-1d90-4696-9678-d7cc4f9c0db3

I'm not completely sure what _should_ happen in that case, but definitely not that. Probably we should treat the user's override as stable regardless of the changes caused by toggling life/death? Since presumably that's _why_ they're overriding it in the first place.

Do you wanna take a stab at this or should I?

RobertChrist:
Huh.

Apologies for not catching that, and nice find!

Yeah, it looks like its iterating through the options when the stone is marked as alive/dead, which is certainly not right.

Stable is nice, because if you mark a bunch of stones as alive/dead, your changes aren't lost. 

I worry that it might not be obvious, though, where all of your overrides are in that case?  Like, you might mark a stone as dead, and not realize that the score being shown by the app is still the new estimation + your original override, depending on how the territory map changes and where you clicked?

If we erase the overrides every time a stone's status changes, then that is far less error prone, and more obvious to a user what's going on.  Users would catch on pretty quickly that they should just mark stones as alive/dead before overriding to begin with, which I think makes sense.  It also gives the users an easy way to clear all of their overrides.  I think I'm leaning towards this option?

Usually when I'm overriding, its not because I'm arguing the life/death of stones, so much as what I think I can get away with against this opponent xD

I'm up for implementing it, but I'm currently on vacation, and won't be back near a non-work-owned computer until the last week of Jan.  I'm happy for you to take a swing at it, if you'd like to before then!

apetresc:
I haven't thought about it too deeply just yet, but my first instinct is just to add the dead/alive toggles to the key into the `estimateOverrides` object, so it's not just a function of the `(x,y)` coordinates, but rather a function of `((x, y), deadOverrideHash)`, where `deadOverrideHash` is some hash of the stones whose life/death status has been overridden by the user.

That way `estimateOverrides` will store a separate set of overrides for each configuration of dead/alive stones, which I think would be the least surprising way we could handle this. If you mess with the territory/influence estimate, then you change the life+death status of some groups, you have to re-mess with the new estimate. Then if you switch the L+D status back and forth, it keeps the two overrides separately.

Does that sound reasonable?

apetresc:
Reviving this one a couple years later, sorry for the enormous gap! On the life/death clash: I went with your instinct to just clear the overrides whenever a stone gets toggled alive/dead. It's the simplest and least surprising of the options, and like you said, it doubles as an easy "clear all". I talked myself out of the `((x,y), deadOverrideHash)` keying I'd floated back then, it was more clever than it needed to be.

I also rebased onto current master and tidied a few things while I was in here: the override state updates immutably now, keyed by a plain `"x,y"`, and cycles through a bounded set of states instead of an ever-growing click counter. Added a unit test for the cycle and an e2e that clicks through it and checks the rendered territory, plus the clear-on-toggle behavior.

Thanks again for this, it's a genuinely nice feature. Merging now.

RobertChrist:
That's awesome to hear!  I'm happy this and the other PRs got merged in, thank you for staying on top of them. And no worries about the amount of time, as a new father, I get how quickly time can disappear : )  @apetresc

PR #942, opened 2023-07-04
Feature/940 game tree options

RobertChrist:
Implements https://github.com/SabakiHQ/Sabaki/issues/940

### DIFF
--- a/e2e/node-menu.spec.js
+++ b/e2e/node-menu.spec.js
@@ -1,0 +1,109 @@
+const {expect} = require('@playwright/test')
+const {test} = require('./fixtures/electron-app')
+const {loadSgfStringAndWait, waitForRender} = require('./helpers')
+
+// Tests for the game-tree node right-click menu (openNodeMenu).
+//
+// That menu is a *native* Electron menu, so it can't be opened or clicked
+// through the DOM. Instead these tests drive the template that openNodeMenu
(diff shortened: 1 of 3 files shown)


apetresc (commented):
Firstly, let me apologize for the obscene amount of time it's taken me to get to this :( Sorry - but I'm here for it now!

This is a great addition, I agree. The only part of it that I take any issue with is adding the "Toggle Show Comments" button to the right-click menu; it implies that this only opens the comment pane for that particular node, since every other entry on that menu is node-specific. It doesn't seem necessary to have that global option there. After all, if the user has figured out how to enable the game tree from the global View menu then they can probably also figure out how to toggle the comments from the same place too.

So I'm tempted to merge this in but with that menu item removed. Any objections, @RobertChrist?

apetresc (commented):
(I would push this adjustment myself but you haven't enabled maintainer access to your branch; so let me know if you want to make it on your end, or I can just merge your commits on a separate branch 🙂)

apetresc (commented):
(Oh actually I see you *do* have "Maintainers are allowed to edit this pull request" on, but I guess Github doesn't recognize me as the maintainer because it belongs to the SabakiHQ org and yishn owns that. Oh well, just let me know if you want to make this adjustment or I'll just push a separate branch tomorrow)

Issue #756, opened 2020-11-23
Beginner suggestion - Engine management - ADD labels


David53bcg:
For go players like me who are not developers, it might be helpful to clarify the two labels within Engine Management/Add, so that they are more informative for Sabaki's wider potential user base. e.g. : 'Path (engine .exe file only)' and 'Arguments (all other). Perhaps make the same point in the user documentation. 

The guidance given in engines' documentation is sometimes generalised and therefore silent on Sabaki's particular two-line setup. It took forum advice from three very patient experts to identify the source of my repeated failure to get one particular engine running correctly - some of which would have been unnecessary if the Add labels had been clearer for beginners.

Thanks for a great SGF editor. It is indeed elegant!

David
London

PR #967, opened 2024-12-17
Make helptext for adding engines clearer for nontechnical users

Chewt:
The helptext for adding a new engine had room for improvement to allow for nontechnical users to better understand how to add an engine to Sabaki.

This was brought up in #756 when the author had trouble adding an engine to Sabaki. This PR hopes to clarify confusion from users who may just be trying to follow command line instructions from Engine documentation and not know much about command lines.

### DIFF
--- a/src/components/drawers/PreferencesDrawer.js
+++ b/src/components/drawers/PreferencesDrawer.js
@@ -666,7 +666,9 @@ class EngineItem extends Component {
         ),
         h('input', {
           type: 'text',
-          placeholder: t('Path'),
+          placeholder: t(
+            'Path (e.g., /path/to/engine or C:\\path\\to\\engine.exe)'
+          ),
           value: path || '',
           name: 'path',
           onChange: this.handleChange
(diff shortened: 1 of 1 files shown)


sente361:
Thanks @Chewt. Many people would find this useful. It would also be very useful to include a guide to choosing the best settings in the gtp line and the time settings line. I use the settings below but I have no idea whether or not they are best for my system.
_gtp -override-config ponderingEnabled=false,maxVisits=200000 -model KataGo20b-1.1.gz
time_settings 0 10 1;_

Chewt:
The only problem with including a guide on the settings, is that they are specific to the engine itself. So those settings you mentioned are specific to katago. We would either need to include a guide that lists settings for all of the common engines, or link to the documentation to the engines themselves. I would lean toward the second option since that documentation already exists.

Chewt:
Actually, I think this pretty much covers the basic usecase to get people started: https://github.com/SabakiHQ/Sabaki/blob/master/docs/guides/engines.md 

Maybe this could somehow be integrated into the GUI somewhere under the help menu, but that's not really related to this PR

PR #1018, opened 2026-06-16
Added keyboard shortcuts for variation-related menu items; also for toggling good/bad move

MarcelGruenauer:
Added keyboard shortcuts for various existing menu items related to editing variations:

- CmdOrCtrl+Alt+Shift+C = Edit > Copy Variation
- CmdOrCtrl+Alt+Shift+X = Edit > Cut Variation
- CmdOrCtrl+Alt+Shift+V = Edit > Paste Variation
- CmdOrCtrl+Alt+Shift+Left = Edit > Make Main Variation
- CmdOrCtrl+Shift+Left = Edit > Shift Left
- CmdOrCtrl+Shift+Right = Edit > Shift Right
- CmdOrCtrl+Shift+Backspace = Edit > Remove Other Variations

Also added menu items for "Toggle Good Move" and "Toggle Bad Move" so we can use keyboard shortcuts for them.

- CmdOrCtrl+Shift+] = Edit > Toggle Good Move
- CmdOrCtrl+Shift+[ = Edit > Toggle Bad Move

### DIFF
--- a/e2e/annotation-toggle.spec.js
+++ b/e2e/annotation-toggle.spec.js
@@ -1,0 +1,62 @@
+const {expect} = require('@playwright/test')
+const {test} = require('./fixtures/electron-app')
+const {loadSgfStringAndWait, waitForRender} = require('./helpers')
+
+// Coverage for sabaki.toggleMoveAnnotation — the helper behind the
+// "Toggle Good Move" / "Toggle Bad Move" menu items (menu.js). Move annotations
+// (TE/IT/DO/BM) are mutually exclusive, so toggling one on must clear any other,
+// and toggling the same one again clears it entirely. The exclusivity itself
+// lives in setComment; this exercises the toggle-on/off + replace behavior end
+// to end.
+
+// Root plus a single Black move, so the current position is a real move node.
+const SGF = '(;GM[1]FF[4]SZ[19];B[pd])'
+
+function moveProp(page, prop) {
+  return page.evaluate((p) => {
+    const s = window.__sabaki
+    const tree = s.state.gameTrees[s.state.gameIndex]
+    const data = tree.get(s.state.treePosition).data
+    return data[p] != null ? data[p] : null
+  }, prop)
+}
+
+async function toggle(page, annotation) {
+  await page.evaluate((a) => {
+    const s = window.__sabaki
+    s.toggleMoveAnnotation(s.state.treePosition, a)
+  }, annotation)
+  await waitForRender(page)
+}
+
+test.describe('toggleMoveAnnotation', () => {
+  test.beforeEach(async ({page}) => {
+    await loadSgfStringAndWait(page, SGF)
+    await page.evaluate(() => window.__sabaki.goToEnd())
+    await waitForRender(page)
+  })
+
+  test('toggles a move annotation on, then off', async ({page}) => {
+    expect(await moveProp(page, 'TE')).toBeNull()
+
+    await toggle(page, 'TE')
+    expect(await moveProp(page, 'TE')).not.toBeNull()
+
+    await toggle(page, 'TE')
+    expect(await moveProp(page, 'TE')).toBeNull()
+  })
+
+  test('replaces an existing annotation instead of stacking them', async ({
+    page,
+  }) => {
+    await toggle(page, 'TE')
+    expect(await moveProp(page, 'TE')).not.toBeNull()
+    expect(await moveProp(page, 'BM')).toBeNull()
+
+    // Toggling Bad Move while Good Move is set must clear TE, not coexist —
+    // a node can never be annotated as both good and bad.
+    await toggle(page, 'BM')
+    expect(await moveProp(page, 'BM')).not.toBeNull()
+    expect(await moveProp(page, 'TE')).toBeNull()
+  })
+})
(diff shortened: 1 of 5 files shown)


apetresc:
I'll admit at first I was skeptical that these would be ergonomic or useful, but I'm pretty sold after trying them out. They seem to naturally fit their function for the most part.

The only one I'm not convinced about is the one that you didn't even document, `CmdOrCtrl+Shift+1` to toggle player. It's not intuitively related to the function like the others are, and I'm not sure why anyone would be toggling players often enough to need a shortcut. Thoughts? Did you even intend to include that one in the PR, given that you didn't mention it in the description?

MarcelGruenauer:
Ah sorry, I must have forgot to document it. I use toggling players, both in play and in edit mode, all the time, because I am entering problems. So when in a problem it is the opponent's turn to play I use that toggle. I wrote some software to take problem SGF trees and generate a series of problems. Among many other indicators, it also uses "tesuji" and "bad move" to determine what to include in the generated problems, so I use those toggles all the time as well.

I'm not familiar with pull requests;  what should I do now to move forward? I have rebased the add-keyboard-shortcut feature branch to the new official master branch, so should I make a new pull request where I also improve the documentation?

Thank you for considering this feature! In my fork I have (vibe) added many other features that I find useful and use all the time; please let me know if you're open to more pull requests.

apetresc:
Oh, don't worry about it, I can do the rebasing and changes. I just wanted to get your thoughts.

You're right that problem entry is a good use case for toggling players. I'm convinced we should keep it. I just want to propose an alternative binding than Cmd-Shift-1: I propose Command-Shift-P. I think it's better for two reasons:

- 'P' for player is an easier mnemonic to remember
- Cmd-P (without Shift) is already "Pass", which also toggles the player, so it makes a lot of sense for Cmd-Shift-P to be a sort of "alternate form" of Cmd-P, which is what Shift-modifiers usually do.

I can make the change, rebase, and merge on my end. I just want to get your approval that this is a good compromise, as the champion of this feature :)

apetresc:
I pushed an updated version straight to this branch so it's easy to try end-to-end: it's your commit rebased onto the current `master`, plus one commit that changes `Toggle Player` from `Cmd/Ctrl+Shift+1` to `Cmd/Ctrl+Shift+P`, the alternative binding proposed above (`P` for Player, and it lines up with the existing `Cmd+Shift+E` / `Cmd+Shift+R` in the Play menu).

Because the rebase rewrites history, your local branch won't fast-forward onto it. To sync your local copy to exactly what's here:

```sh
# in your clone of the fork
git fetch origin
git checkout add-keyboard-shortcuts
git reset --hard origin/add-keyboard-shortcuts
```

If you have any unpushed local commits on this branch you want to keep, back them up first (`git branch backup-add-keyboard-shortcuts`) before the reset.

Then build it and give it a spin, especially `Cmd+Shift+P` for Toggle Player in your problem-entry workflow, and let me know how it feels. If it works well for you, then this should be good to go.

Thanks again for the contribution!

apetresc:
Ah, no worries — looks like that last push replaced the version I'd pushed instead of building on top of it. The upside is it rebased cleanly onto the latest master (which now has the macOS quit fix in it too), but it dropped the small commit I'd added, so `Toggle Player` is back on `Cmd+Shift+1`.

Getting `Cmd+Shift+P` back is a one-liner. In `src/menu.js`, on the `Toggle Player` item, change:

```js
accelerator: 'CmdOrCtrl+Shift+1',
```

to:

```js
accelerator: 'CmdOrCtrl+Shift+P',
```

Then just commit and push to this branch normally — no force-push needed this time, since it's a new commit on top rather than a history rewrite. (Or if it's easier, say the word and I'll pop it back on for you.)

After that I think we're basically there. Thanks for bearing with the git gymnastics!

MarcelGruenauer:
Yes, Cmd+Shift+P  works as well. (I chose "1" because it's like Cmd-1 in edit mode). Thank you!

apetresc:
Since you'd already okayed Cmd-Shift-P, I just knocked the rest out on my end. Two commits on top of yours, both fast-forwards this time so your commit's untouched:

- Cmd-Shift-P for Toggle Player, plus a Prettier pass so `menu.js` passes the format-check CI step (the Good/Bad Move handlers were failing it on formatting).
- Refactored the Good/Bad Move toggles a little. Sabaki already had this "toggle one of the mutually-exclusive move annotations" logic in `getCommentMenuTemplate`, and `setComment` already guarantees only one of TE/IT/DO/BM sticks, so I pulled it into a small `sabaki.toggleMoveAnnotation(pos, 'TE'|'BM')` helper and pointed your two menu items at it. Keeps `menu.js` out of the game-tree internals and reuses the exclusivity we already had. Added a quick e2e test for it too.

Your branch will just fast-forward this time, so a normal pull does it, no reset:

```sh
git checkout add-keyboard-shortcuts
git pull origin add-keyboard-shortcuts
```

I think that's basically it. Give it a build to make sure I didn't wreck your workflow, but should be good to merge. Thanks again, and happy to take a look at the other stuff you mentioned if you want to send it over.

MarcelGruenauer:
I built it; it looks good, thank you!

PR #1022, opened 2026-06-20
Add option to show score/winrate change instead of absolute value

### LINKED ISSUE #843: Option to show score difference instead of score lead on heat map
During engine analysis, the heat map shows the absolute score lead of the moves being analyzed. I propose to add an option to display the score difference instead of the score lead (i.e. what OGS does). For example, the best move would usually be around +0 and worse moves would have a negative value showing the number of points lost by playing that move.

kevinsung:
Fixes #843 

- Adds a new View menu "Analysis Value Display" with options "Absolute" and "Change" to pick whether to show the absolute winrate/scorelead or the change from the previous value.
- Adds a keyboard shortcut Ctrl+Shift+V to toggle between these two settings (V stands for "Value").
- Factors the logic for determining what value to display out of `Goban.js` into `getAnalysisHeatMapCell` in `analysis.js` to enable unit testing.

Disclaimer: I used Claude Code to write the tests.

image

### DIFF
--- a/e2e/analysis-value-display.spec.js
+++ b/e2e/analysis-value-display.spec.js
@@ -1,0 +1,134 @@
+const {expect} = require('@playwright/test')
+const path = require('path')
+const {test} = require('./fixtures/electron-app')
+const {
+  loadSgfAndWait,
+  attachAndWaitForEngines,
+  detachAndWait,
+  waitForRender,
+} = require('./helpers')
+
+// End-to-end coverage of the "Analysis Value Display" setting
+// (board.analysis_value_type, menu.js / Goban.js): 'absolute' shows each
+// candidate move's raw winrate, while 'change' shows its difference from the
+// best move, so the best move always reads as 0 and every other move reads
+// as its cost relative to it.
+//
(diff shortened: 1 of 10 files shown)


apetresc:
Thanks for this, nice addition! I just made a couple of small maintainer edits on top:

  - Rebased onto latest `master`, there's been a bit of churn lately.
  - **Folded the picker into the existing "Analysis Metric" submenu** (under a separator) instead of a separate top-level "Analysis Value Display" submenu. Not only does it keep the top-level "View" menu cleaner, but it also just helps explain what "Analysis Value Display" is actually referring to. This is just overall less confusing, imo.
pr-1022-analysis-metric-menu

  - Renamed all your instances of "change" to "relative", I think that's the more standard terminology in analysis circles.

  Unit and e2e tests pass on the rebased branch.
  
  Any comments on my changes, have I borked anything? Happy to merge once you okay this.

apetresc:
(Pinging @kevinsung )

kevinsung:
Those changes are great and it works for me! Thank you, please go head and merge.

PR #1044, opened 2026-07-06
fix: prevent argv flags and app path from being treated as open file

### LINKED ISSUE #954: "This file is unreadable" message on start up
When I open Sabaki 0.52.2 I get the following error message: "This file is unreadable". It's not terrible since I can just click ok and use it normally, but it's annoying to get an error message every time the program is launched

I'm using ubuntu 22.04 and I've installed sabaki via snap (2.61.1).

DeepChirp:
The old argv parsing checked whether process.argv[0] ended with "electron" or "electron.exe" to distinguish dev mode from packaged mode. On Linux distributions where the Electron binary can be named electron42, electron32, etc., this check fails, causing the code to fall into the wrong branch and treat the app.asar path or Chromium flags (--no-sandbox, --disable-gpu, etc.) as a file to open.

Replace the fragile binary-name check with robust content-based filtering: filter out internal --flags, as well as the app entry paths ("." for dev mode, or ".asar" for packaged mode). The first remaining argument (if any) is the actual file to open.

Fix #954

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -12,6 +12,7 @@ const {resolve} = require('path')
 const i18n = require('./i18n')
 const setting = require('./setting')
 const updater = require('./updater')
+const {getOpenFileFromArgv} = require('./modules/utils')
 
 let windows = []
 let openfile = null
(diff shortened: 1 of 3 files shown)


apetresc:
Thanks @DeepChirp -- good catch on the binary-name heuristic, that was overdue. I pushed a small maintainer edit on top: pulled the filter out into a pure `getOpenFileFromArgv()` in `utils.js` with unit tests over the launch matrix (dev, renamed Linux binary, packaged win/linux/mac, `--no-sandbox` injection, `.asar` entry), and widened the filter from `--` to any leading `-` so a stray single-dash arg like macOS `-psn_…` can't be taken as a file either. `npm test` + the e2e are green. Merging -- appreciate the fix!

PR #1063, opened 2026-07-12
Include variation count into theme manifest

lykahb:
This gives users and theme authors more control over board presentation.

Here is a motivation for the changes - a theme with many variations per color and boards that include grid into the image.
image

Themes can now declare different numbers of black and white stone variations through `stoneVariations` in `package.json`. Sabaki translates the manifest’s color names into Shudan’s numeric signs and passes the counts to the renderer. Validation and the default count remain in Shudan so the renderer has one authoritative policy and remains safe for non-Sabaki consumers.

Theme selection is also propagated through application state, ensuring variation counts update immediately when switching themes.

Another change is that custom background images selected in Sabaki’s settings now can override the background included in theme. The css rule now has higher specificity. I can extract that into another PR if you prefer.

Requires the accompanying https://github.com/SabakiHQ/Shudan/pull/31

### DIFF
--- a/docs/guides/create-themes.md
+++ b/docs/guides/create-themes.md
@@ -24,6 +24,8 @@ will only use the following fields:
 - `author` _(optional)_
 - `homepage` _(optional)_
 - `main` - The CSS file to include in Sabaki, usually `styles.css`
+- `stoneVariations` _(optional)_ - Positive integer variation counts for black
+  and white stones; each color defaults to five
 
 For example:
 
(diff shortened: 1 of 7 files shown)


apetresc:
Thanks for this, the variation-count feature is a nice addition! One request before I dig into the review properly: could you split the ThemeManager background-specificity change (the `main` -> `#main main` selector) out into its own PR? It's really a separate concern from the stone variations, and I'd rather land it on its own so each change keeps a clean history. The rest I'll review together with the Shudan side (#31). Thanks!

lykahb:
@apetresc Sure, extracted it into https://github.com/SabakiHQ/Sabaki/pull/1073.

apetresc:
Shudan 1.8.0 is out on npm now (it includes your #31), so I rebased this onto current master and bumped `@sabaki/shudan` to ^1.8.0. The branch is pointed at the released version now, and the full e2e suite is green here with no regressions.

Could you give it a run end to end with a real theme? Drop a `stoneVariations` block into a theme's package.json (say `{"black": 12, "white": 7}`), point Sabaki at it, and confirm the extra variation classes actually land on the stones. Mainly want to be sure the manifest-to-Shudan wiring behaves the way you intended before this goes in. Thanks for the nice work on this!

lykahb:
Thanks! I pulled the branch and tested it with 24 variations per stone color, using a real WIP theme. The random classes get assigned as expected

